PR #1, opened 2024-06-28
compliance: compliance-server: Define initial compliance api

joeykraut:
### Purpose
This PR adds the initial skeleton for the compliance API. This defines API types and a basic no-op API server for the compliance route.

### Todo
- Define database schemas + migration
- Use db + chainlink API to assess wallet risk

### Testing
- Manually tested the server

### DIFF
--- a/compliance/compliance-api/src/lib.rs
+++ b/compliance/compliance-api/src/lib.rs
@@ -1,0 +1,17 @@
+use serde::{Deserialize, Serialize};
+
+/// The response type for a compliance check
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ComplianceCheckResponse {
+    /// The compliance status of the wallet
+    pub compliance_status: ComplianceStatus,
+}
+
+/// The status on compliance for a wallet
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum ComplianceStatus {
+    /// The wallet is compliant
+    Compliant,
+    /// The wallet is not compliant
+    NotCompliant,
+}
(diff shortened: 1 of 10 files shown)


compliance/compliance-api/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+use serde::{Deserialize, Serialize};
+
+/// The request type for a compliance check
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ComplianceCheckRequest {
+    /// The wallet id to request
+    pub wallet_id: String,
```
akirillo, line 7: Should this be an address instead of a wallet ID? Also, are we going to be using this request type, or are we going to be parsing the wallet address from the route (as you have it currently configured)?

compliance/compliance-server/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+#[tokio::main]
+async fn main() {
+    setup_system_logger(LevelFilter::INFO);
+    let cli = Cli::parse();
+
+    // Get compliance information for a wallet
+    let chainalysis_key = cli.chainalysis_api_key.clone();
+    let compliance_check = warp::get()
```
akirillo, line 28: just curious, why did we choose to use `warp` here instead of `hyper` (as we do in the relayer)?

akirillo (approved):

compliance/compliance-api/src/lib.rs
joeykraut: oops yes, thanks

compliance/compliance-server/src/main.rs
joeykraut: we don't need the low level interface hyper gives us here. In the relayer it's helpful for more custom auth flows and routing, but here we just want something simple -- which warp gives us

PR #3, opened 2024-06-29
compliance: compliance-server: Fallback to chainalysis API if not cached

joeykraut:
### Purpose
This PR implements the fallback to using the chainalysis API in the case that a wallet's risk assessment is not cached in the DB. If chainalysis delivers a successful risk assessment, we then insert it into our DB.

### Testing
- Tested with my wallet addresses (compliant)
- Tested with a known sanctioned address (non-compliant)

### DIFF
--- a/compliance/compliance-server/migrations/2024-06-28-231756_create_compliance_table/up.sql
+++ b/compliance/compliance-server/migrations/2024-06-28-231756_create_compliance_table/up.sql
@@ -2,6 +2,7 @@
 CREATE TABLE IF NOT EXISTS wallet_compliance (
     address TEXT PRIMARY KEY,
     is_compliant BOOLEAN NOT NULL,
+    risk_level TEXT NOT NULL,
     reason TEXT NOT NULL,
     created_at TIMESTAMP NOT NULL DEFAULT NOW(),
     expires_at TIMESTAMP NOT NULL DEFAULT NOW() + INTERVAL '1 year'
--- a/compliance/compliance-server/src/db.rs
+++ b/compliance/compliance-server/src/db.rs
@@ -1,6 +1,6 @@
 //! Database helpers for the server
 
-use std::time::SystemTime;
+use std::time::{Duration, SystemTime};
 
 use compliance_api::ComplianceStatus;
 use diesel::{ExpressionMethods, Insertable, PgConnection, QueryDsl, Queryable, RunQueryDsl};
@@ -14,6 +14,9 @@ use crate::{
     },
 };
 
+/// The default expiration duration for a compliance entry
+const DEFAULT_EXPIRATION_DURATION: Duration = Duration::from_days(365);
+
 // ----------
 // | Models |
 // ----------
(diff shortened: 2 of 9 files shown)


compliance/compliance-server/migrations/2024-06-28-231756_create_compliance_table/up.sql
```diff
@@ -2,6 +2,7 @@
 CREATE TABLE IF NOT EXISTS wallet_compliance (
     address TEXT PRIMARY KEY,
     is_compliant BOOLEAN NOT NULL,
+    risk_level TEXT NOT NULL,
```
akirillo, line 5: is it safe to edit migrations like this 😅 or should we just make a new one? though I get that nothing is in prod yet so its chill

compliance/compliance-server/src/db.rs
```diff
@@ -25,12 +28,20 @@ use crate::{
...
     pub is_compliant: bool,
+    pub risk_level: String,
     pub reason: String,
     pub created_at: SystemTime,
     pub expires_at: SystemTime,
 }
 
 impl ComplianceEntry {
+    /// Create a new entry from a risk assessment
+    pub fn new(address: String, is_compliant: bool, risk_level: String, reason: String) -> Self {
```
akirillo, line 39: nit: why not have this accept a `RiskAssessmentResponse` directly? will read more nicely

compliance/compliance-server/src/main.rs
```diff
@@ -102,5 +107,11 @@ async fn check_wallet_compliance(
     }
 
     // 2. If not present, check the chainalysis API
-    Ok(ComplianceStatus::Compliant)
+    info!("address not cached in DB, querying Chainalysis");
+    let compliance_status = query_chainalysis(&wallet_address, chainalysis_api_key).await?;
+    let compliance_entry = compliance_status.as_compliance_entry();
```
akirillo, line 112: nit: i'm sure this is in the next PR, but `query_chainalysis` will return a `RiskAssessmentResponse`, not a `ComplianceStatus`. I think it makes sense for it to return a `ComplianceEntry`, as this code path accomplishes, by calling `ComplianceEntry::new(resp)` in `query_chainalysis`

akirillo (approved):

compliance/compliance-server/migrations/2024-06-28-231756_create_compliance_table/up.sql
joeykraut: no it is not safe, as you mention this is only done because nothign is deployed

compliance/compliance-server/src/db.rs
joeykraut: separation of concerns; we should be able to create a `ComplianceEntry` without needing to construct a chainalysis response. I see no reason to couple them.

compliance/compliance-server/src/main.rs
joeykraut: sure that works

PR #4, opened 2024-06-29
compliance: compliance-server: Add Dockerfile

joeykraut:
### Purpose
This PR adds a `Dockerfile` for the compliance server to deploy it on an ECS cluster. As well, I added env variable fallbacks for the secrets used to configure the compliance server. We will configure these via secrets manager.

### Testing
- Built and ran the dockerfile and tested the compliance server

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*.log
+.env
+**/Dockerfile
--- a/compliance/compliance-server/Dockerfile
+++ b/compliance/compliance-server/Dockerfile
@@ -1,0 +1,46 @@
+FROM --platform=arm64 rust:latest AS chef
+
+# Create a build dir and add local dependencies
+WORKDIR /build
+
+COPY ./rust-toolchain ./rust-toolchain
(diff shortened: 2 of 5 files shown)


.dockerignore
```diff
@@ -0,0 +1,4 @@
+node_modules
```
akirillo, line 1: lol, not that it matters, but we'll never have a `node_modules` in here anyway right

compliance/compliance-server/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# Disable compiler warnings and enable backtraces for panic debugging
+ENV RUSTFLAGS=-Awarnings
+ENV RUST_BACKTRACE=1
+
+# Build only the dependencies to cache them in this layer
+RUN cargo chef cook --release --recipe-path recipe.json
+
+# Build the compliance server
+COPY . .
+RUN cargo build --release
```
akirillo, line 34: assuming the build context is the root of this repo, should we `cargo build --release --package compliance-server`?

akirillo (approved):

PR #5, opened 2024-07-17
funds-manager: Refactor into HTTP server setup

joeykraut:
### Purpose
This PR changes the runtime of the `fee-sweeper` to be http server based. I also generalized this into a `funds-manager` as we will add on a custody client and functionality to move funds to/from storage and hot wallets.

### Todo
- Refactor the diesel queries to be async so that we can match the `warp` interface
- Define API routes

### Testing
Deferred until the todos are done


funds-manager/src/indexer/queries.rs
```diff
@@ -56,6 +56,9 @@ pub(crate) struct FeeValue {
     /// The mint of the fee
     #[sql_type = "Text"]
     pub mint: String,
+    /// The receiver of the mint
+    #[sql_type = "Text"]
+    pub receiver: String,
```
akirillo, line 61: does this require a migration? not seeing a new one in this PR

akirillo (approved):

funds-manager/src/indexer/queries.rs
joeykraut: no, this is a query-specific struct, so it defines the return type expected by a specific query. you can see down below that i added the `receiver` field to the `SELECT` statement

PR #6, opened 2024-07-18
funds-manager: Use `diesel-async` and add `warp` endpoints

joeykraut:
### Purpose
This PR refactors the `funds-manager` to use `diesel-async` which allows us to fit the funds manager endpoints into the `warp` interface. I expose two endpoints `index-fees` and `redeem-fees`.

### Todo
- Move the endpoints and request types into an API crate as in other extensions

### Testing
- Tested both endpoints

### DIFF
--- a/funds-manager/src/db/mod.rs
+++ b/funds-manager/src/db/mod.rs
@@ -1,5 +1,37 @@
 //! Database code
 
+use diesel_async::AsyncPgConnection;
+use native_tls::TlsConnector;
+use postgres_native_tls::MakeTlsConnector;
+use renegade_util::err_str;
+use tracing::error;
+
+use crate::error::FundsManagerError;
+
 pub mod models;
 #[allow(missing_docs)]
 pub mod schema;
+
+/// Establish a connection to the database
+pub async fn establish_connection(db_url: &str) -> Result<AsyncPgConnection, FundsManagerError> {
+    // Build a TLS connector, we don't validate certificates for simplicity.
+    // Practically this is unnecessary because we will be limiting our traffic to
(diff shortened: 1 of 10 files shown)


funds-manager/src/db/mod.rs
```diff
@@ -1,5 +1,36 @@
...
+        .map_err(err_str!(FundsManagerError::Db))?;
+    let connector = MakeTlsConnector::new(connector);
+    let (client, conn) = tokio_postgres::connect(db_url, connector)
+        .await
+        .map_err(err_str!(FundsManagerError::Db))?;
+
+    // Spawn the connection handle in a separate task
+    tokio::spawn(async move {
+        if let Err(e) = conn.await {
+            eprintln!("Connection error: {}", e);
```
akirillo, line 31: not the most pressing, but a thought for down the road: would be nice to use `tracing`'s logging features so that we can get structured logging that a datadog sidecar can pick up

funds-manager/src/main.rs
```diff
@@ -161,13 +164,81 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+        .await
+        .map_err(|e| warp::reject::custom(ApiError::RedemptionError(e.to_string())))?;
+    Ok(warp::reply::json(&"Fees redeemed successfully"))
+}
+
+// -----------
+// | Helpers |
+// -----------
+
+/// Handle a rejection from an
```
akirillo, line 223: nit:
```suggestion
/// Handle a rejection from a route
```
?

akirillo (approved):

PR #7, opened 2024-07-19
bin, funds-manager: Add deployment and upgrade scripts

joeykraut:
### Purpose
This PR adds a `Dockerfile` to the funds manager and generalizes the existing `build_and_push.sh` and `deploy.sh` scripts to work for all extensions.

 ### Testing
- Deployed to testnet


bin/build_and_push.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+REGION=us-east-2
```
akirillo, line 2: this should probably be configurable via command line args as well for e.g. price reporter extension

bin/deploy.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
+set -e
+
+REGION=us-east-2
```
akirillo, line 4: same here, region should prob be configurable via CLI

bin/deploy.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Derive values from environment and resource
+CLUSTER_NAME="$ENVIRONMENT-$RESOURCE-cluster"
+SERVICE_NAME="$ENVIRONMENT-$RESOURCE-service"
+TASK_DEF_NAME="$ENVIRONMENT-$RESOURCE-task-def"
+ECR_REPO="$RESOURCE-$ENVIRONMENT"
+
+# Fetch the latest image URI from ECR
+ECR_URL="$ECR_REGISTRY/$ECR_REPO"
+IMAGE_URI=$(aws ecr describe-images --repository-name $ECR_REPO --region $REGION --query 'sort_by(imageDetails,& imagePushedAt)[-1].imageTags[0]' --output text)
```
akirillo, line 31: empirically, this has often given me the `:latest`-tagged image URI, as opposed to the `COMMIT_HASH`-tagged image URI. This is unsafe to put in the task definition as new tasks getting started will use whatever image is tagged w/ latest at launch time. I'd propose we just generate `COMMIT_HASH` as we do in `build_and_push.sh` and use that to construct the `FULL_IMAGE_URI`

akirillo (approved):

bin/deploy.sh
joeykraut: done

PR #8, opened 2024-07-20
funds-manager: Use dummy key for arbitrum client

joeykraut:
### Purpose
This PR updates the CLI in two ways:
1. Allows each config entry to be set by an environment variable. This will make it easier to deploy the `funds-manager` via terraform.
2. Removes the Arbitrum private key argument. We don't need a private key for any arbitrum client operations, so instead we use a dummy key -- the master key used in nitro dev nodes.

### Testing
- Ran the funds manager with the dummy key and environment variables

### DIFF
--- a/funds-manager/Dockerfile
+++ b/funds-manager/Dockerfile
@@ -51,4 +51,4 @@ RUN apt-get update && \
 COPY --from=builder /build/target/release/funds-manager /bin/funds-manager
 
 ENTRYPOINT ["/bin/funds-manager"]
-CMD ["--help"]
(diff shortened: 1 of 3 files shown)


akirillo (commented):
Won't we need a keypair soon enough to manage a hot wallet? can just leave the CLI arg and default to this for devex

joeykraut:
> Won't we need a keypair soon enough to manage a hot wallet? can just leave the CLI arg and default to this for devex

We can add it back when we need it

akirillo (approved):

PR #11, opened 2024-07-23
funds-manager: custody-client: Add withdrawal handler

joeykraut:
### Purpose
This PR adds an endpoint handler for the withdrawal path in the `funds-manager`'s custody client. This endpoint will find the correct vault and asset to withdraw from (borrowing code from the deposit flow) then initiate a withdrawal through Fireblocks.

### Todo
- Authentication for custody API

### Testing
- Deposited a few ERC20s into Fireblocks, withdrew into a personal wallet through the endpoint

### DIFF
--- a/funds-manager/funds-manager-server/src/custody_client/mod.rs
+++ b/funds-manager/funds-manager-server/src/custody_client/mod.rs
@@ -1,14 +1,21 @@
 //! Manages the custody backend for the funds manager
 #![allow(missing_docs)]
 pub mod deposit;
+pub mod withdraw;
 
 use ethers::prelude::abigen;
 use ethers::providers::{Http, Provider};
 use ethers::types::Address;
-use fireblocks_sdk::{Client as FireblocksClient, ClientBuilder as FireblocksClientBuilder};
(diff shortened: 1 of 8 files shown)


funds-manager/funds-manager-server/src/custody_client/mod.rs
```diff
@@ -20,20 +27,24 @@ abigen!(
...
+#[derive(Clone, Copy)]
+pub(crate) enum DepositWithdrawSource {
+    /// A Renegade quoter deposit or withdrawal
     Quoter,
-    /// A fee withdrawal
-    FeeWithdrawal,
+    /// A fee redemption deposit
+    FeeRedemption,
+    /// A gas withdrawal
+    RelayerGas,
```
akirillo, line 37: nit: we'll also need gas for the quoters (for permit2 approvals & deposits to custody) so we may wanna rename this

akirillo (approved):

funds-manager/funds-manager-server/src/custody_client/mod.rs
joeykraut: good call

PR #13, opened 2024-07-25
funds-manager: Add gas refuel api

joeykraut:
### Purpose
This PR adds an API endpoint for refueling gas. We segment the vaults from which gas and quoter funds are drawn to more accurately track and allocate our custodied funds.

### Testing
- Hit the endpoint to withdraw gas to a personal wallet

### DIFF
--- a/funds-manager/funds-manager-server/src/handlers.rs
+++ b/funds-manager/funds-manager-server/src/handlers.rs
@@ -3,13 +3,17 @@
 use crate::custody_client::DepositWithdrawSource;
 use crate::error::ApiError;
 use crate::Server;
-use funds_manager_api::{DepositAddressResponse, WithdrawFundsRequest};
+use funds_manager_api::{DepositAddressResponse, WithdrawFundsRequest, WithdrawGasRequest};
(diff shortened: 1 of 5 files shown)


funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -3,13 +3,17 @@
...
 use std::collections::HashMap;
 use std::sync::Arc;
 use warp::reply::Json;
 
 /// The "mint" query param
 pub const MINT_QUERY_PARAM: &str = "mint";
+/// The asset used for gas (ETH)
+pub const GAS_ASSET_NAME: &str = "ETH";
+/// The maximum amount of gas that can be withdrawn at a given time
+pub const MAX_GAS_WITHDRAWAL_AMOUNT: f64 = 0.1; // 0.1 ETH
```
akirillo, line 16: don't mind having this as a constant here, but: 1) we should also make sure this limit is set in FB, and 2) this seems to be unused

akirillo (approved):

PR #16, opened 2024-07-26
funds-manager: fee-indexer: Allow withdrawal of fees to custody

joeykraut:
### Purpose
This PR adds an endpoint to withdraw redeemed fees from fee collector Renegade wallets into our custody stack. This is brokered by the relayer which manages the fee collection wallets.

### Testing
- Redeemed a number of fees and then withdrew them into Fireblocks

### DIFF
--- a/funds-manager/funds-manager-server/src/fee_indexer/fee_balances.rs
+++ b/funds-manager/funds-manager-server/src/fee_indexer/fee_balances.rs
@@ -1,13 +1,32 @@
 //! Fetch the balances of redeemed fees
 
+use crate::custody_client::DepositWithdrawSource;
 use crate::db::models::WalletMetadata;
 use crate::error::FundsManagerError;
-use renegade_api::types::ApiWallet;
-use renegade_common::types::wallet::derivation::derive_wallet_keychain;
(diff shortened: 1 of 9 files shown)


funds-manager/funds-manager-server/src/fee_indexer/fee_balances.rs
```diff
@@ -38,8 +87,88 @@ impl Indexer {
...
 
-        // Fetch the wallet from the relayer
-        let wallet = self.relayer_client.get_wallet(wallet_metadata.id, &root_key).await?;
-        Ok(wallet.wallet)
+        // Fetch the wallet from the relayer and replace the keychain so that we have
+        // access to the full set of secret keys
+        let mut wallet =
+            self.relayer_client.get_wallet(wallet_metadata.id, &root_key).await?.wallet;
+        let keychain =
+            derive_wallet_keychain(&eth_key, self.chain_id).map_err(FundsManagerError::custom)?;
```
akirillo, line 95: do we need to re-derive the keychain here? we can just use the one derived on line 86

akirillo (approved):

PR #19, opened 2024-07-29
funds-manager: Add endpoint to create new hot wallets

joeykraut:
### Purpose
This PR adds an endpoint to create a new hot wallet. Each hot wallet backs a singular vault for which it "caches" funds. We map hot wallets to vaults in a 1:1 manner to properly segment funds by their intended use. 

The hot wallet metadata is stored in the `FundsManager` database, and the private key is stored in AWS secrets manager.

### Testing
- Created hot wallets for each of our existing vaults

### DIFF
--- a/funds-manager/funds-manager-server/src/custody_client/hot_wallets.rs
+++ b/funds-manager/funds-manager-server/src/custody_client/hot_wallets.rs
@@ -1,0 +1,43 @@
+//! Handlers for managing hot wallets
+//!
+//! We store funds in hot wallets to prevent excessive in/out-flow from
+//! Fireblocks
+
+use ethers::{
+    signers::{LocalWallet, Signer},
+    utils::hex::ToHexExt,
+};
+use rand::thread_rng;
+use tracing::info;
+
+use super::CustodyClient;
+use crate::{error::FundsManagerError, helpers::create_secrets_manager_entry_with_description};
+
+impl CustodyClient {
+    /// Create a new hot wallet
+    ///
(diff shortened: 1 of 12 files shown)


funds-manager/funds-manager-server/src/custody_client/hot_wallets.rs
```diff
@@ -0,0 +1,42 @@
...
+        let private_key = wallet.signer().to_bytes();
+
+        // Store the private key in Secrets Manager
+        let secret_name = format!("hot-wallet-{}", address);
+        let secret_value = hex::encode(private_key);
+        create_secrets_manager_entry_with_description(
+            &secret_name,
+            &secret_value,
+            &self.aws_config,
+            "Hot wallet private key",
```
akirillo, line 33: nit: can we have this description (and perhaps the secret name) include which FB vault this hot wallet is for? is very helpful for discoverability if we ever need to manually find the secret

funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -129,3 +129,18 @@ pub(crate) async fn withdraw_fee_balance_handler(
 
     Ok(warp::reply::json(&"Fee withdrawal initiated..."))
 }
+
+/// Handler for creating a hot wallet
+pub(crate) async fn create_hot_wallet_handler(
```
akirillo, line 134: should we have this handler check against a hot wallet already existing in the DB for the given vault? allowing this may confuse subsequent lookup logic

akirillo (approved):

PR #20, opened 2024-07-30
funds-manager: Add endpoint to fetch token balances for hot wallets

joeykraut:
### Purpose
This PR adds an endpoint to query the token balances for hot wallets on-chain. The endpoint fetches all hot wallets and takes the desired mints as query parameters and looks them up on their ERC20 contracts.

### Testing
- Tested the endpoint locally against a hot wallet I funded with a dummy erc20

### DIFF
--- a/funds-manager/funds-manager-server/src/handlers.rs
+++ b/funds-manager/funds-manager-server/src/handlers.rs
@@ -6,14 +6,15 @@ use crate::Server;
 use bytes::Bytes;
 use funds_manager_api::{
     CreateHotWalletRequest, CreateHotWalletResponse, DepositAddressResponse, FeeWalletsResponse,
-    WithdrawFeeBalanceRequest, WithdrawFundsRequest, WithdrawGasRequest,
+    HotWalletBalancesResponse, WithdrawFeeBalanceRequest, WithdrawFundsRequest, WithdrawGasRequest,
 };
+use itertools::Itertools;
 use std::collections::HashMap;
 use std::sync::Arc;
 use warp::reply::Json;
 
-/// The "mint" query param
-pub const MINT_QUERY_PARAM: &str = "mint";
+/// The "mints" query param
+pub const MINTS_QUERY_PARAM: &str = "mints";
 /// The asset used for gas (ETH)
 pub const GAS_ASSET_NAME: &str = "ETH";
 /// The maximum amount of gas that can be withdrawn at a given time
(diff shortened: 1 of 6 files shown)


funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -67,8 +68,8 @@ pub(crate) async fn get_deposit_address_handler(
     query_params: HashMap<String, String>,
     server: Arc<Server>,
 ) -> Result<Json, warp::Rejection> {
-    let mint = query_params.get(MINT_QUERY_PARAM).ok_or_else(|| {
-        warp::reject::custom(ApiError::BadRequest("Missing 'mint' query parameter".to_string()))
+    let mint = query_params.get(MINTS_QUERY_PARAM).ok_or_else(|| {
+        warp::reject::custom(ApiError::BadRequest("Missing 'mints' query parameter".to_string()))
```
akirillo, line 72: pedantic nit: I would keep this parameter as the singular `mint` to make the API more readable (at the expense of an extra constant)

funds-manager/funds-manager-server/src/main.rs
```diff
@@ -342,13 +342,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-        .and(warp::path(CREATE_HOT_WALLET_ROUTE))
         .and(with_hmac_auth(server.clone()))
         .map(with_json_body::<CreateHotWalletRequest>)
         .and_then(identity)
         .and(with_server(server.clone()))
         .and_then(create_hot_wallet_handler);
 
+    let get_hot_wallet_balances = warp::get()
+        .and(warp::path("hot-wallets"))
+        .and(with_hmac_auth(server.clone()))
```
akirillo, line 353: better to be on the safe side, but does this endpoint need hmac auth? i thought the idea was just to protect the "write" / post endpoints

akirillo (approved):

funds-manager/funds-manager-server/src/main.rs
joeykraut: without hmac auth anyone would be able to view our balances, that's certainly not okay

funds-manager/funds-manager-server/src/handlers.rs
joeykraut: disagree that it's more readable, i'll keep it this way. this is a very common pattern to accept list arguments

PR #24, opened 2024-07-31
funds-manager: Lookup Fireblocks asset IDs via ERC20 address

joeykraut:
### Purpose
This PR changes the way we resolve asset IDs in the Fireblocks client. Specifically, we lookup assets directly via the `/supported-assets` API, filter them by contract address, and use the asset ID in that API response.

Previously, we were using the ERC20 symbol name to find asset IDs, which plays poorly with the testnet setup.

### Testing
- Tested both the `/custody/hot-wallets/withdraw-to-hot-wallet` and `/custody/hot-wallets/transfer-to-vault` routes for functionality. These are the two routes that resolve asset IDs directly

### DIFF
--- a/funds-manager/funds-manager-server/src/custody_client/withdraw.rs
+++ b/funds-manager/funds-manager-server/src/custody_client/withdraw.rs
@@ -45,7 +45,7 @@ impl CustodyClient {
         &self,
         source: DepositWithdrawSource,
         destination_address: &str,
-        symbol: &str,
+        mint: &str,
         amount: f64,
     ) -> Result<(), FundsManagerError> {
         let client = self.get_fireblocks_client()?;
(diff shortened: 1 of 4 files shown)


funds-manager/funds-manager-server/src/custody_client/withdraw.rs
```diff
@@ -55,30 +55,36 @@ impl CustodyClient {
...
+            FundsManagerError::Custom(format!("Asset not found for mint: {mint}"))
         })?;
 
         // Check if the available balance is sufficient
+        let available = vault
+            .assets
+            .iter()
+            .find(|a| a.id == asset_id)
+            .map(|acct| acct.available.clone())
+            .unwrap_or_default();
```
akirillo, line 68: are we sure we want to `unwrap_or_default` here? that would result in the "Insufficient balance" error whereas a clearer one would be something to the tune of `FundsManagerError::Custom("Asset not found in vault")`

akirillo (approved):

funds-manager/funds-manager-server/src/custody_client/withdraw.rs
joeykraut: those are the same thing

PR #25, opened 2024-08-01
funds-manager: Add internal wallet id mapping & use internal transaction

joeykraut:
### Purpose
This PR makes the following changes to bring the `funds-manager` into a working state with our testnet environment:
- Remove the sandbox flag on the Fireblocks client
- Add an `internal_wallet_id` mapping to the hot wallets table. Fireblocks requires that we specify whitelisted withdrawals to a wallet ID rather than its address. I manually added wallet IDs for each of the existing hot wallets
- Use a `create_transaction_peer` with `PeerType::INTERNAL_WALLET` to withdraw from Fireblocks. This allows us to control where our funds flow to via a TAP

### Testing
- Tested withdrawing from fireblocks and depositing into it

### DIFF
--- a/funds-manager/funds-manager-api/src/lib.rs
+++ b/funds-manager/funds-manager-api/src/lib.rs
@@ -96,6 +96,8 @@ pub struct WithdrawGasRequest {
 pub struct CreateHotWalletRequest {
     /// The name of the vault backing the hot wallet
     pub vault: String,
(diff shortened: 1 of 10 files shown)


akirillo (approved):
LGTM, does this need a migration due to the new col?

joeykraut:
> LGTM, does this need a migration due to the new col?

forgot to commit it, thanks

PR #28, opened 2024-08-02
funds-manager: custody-client: Add endpoint for registering a gas wallet

joeykraut:
### Purpose
This PR adds an endpoint for a peer to register a gas wallet assigned to them. The funds manager will then mark the wallet as active and assigned to the given peer, fetch the wallet's private key from secrets manager, and return the key to the client.

### Testing
- Hit the endpoint, checked the database entries & the key
- Hit the endpoint enough times to consume all available keys, verified that I received an error

### DIFF
--- a/funds-manager/funds-manager-server/src/custody_client/queries.rs
+++ b/funds-manager/funds-manager-server/src/custody_client/queries.rs
@@ -27,6 +27,33 @@ impl CustodyClient {
         Ok(())
     }
 
+    /// Find an inactive gas wallet
+    pub async fn find_inactive_gas_wallet(&self) -> Result<GasWallet, FundsManagerError> {
+        let mut conn = self.get_db_conn().await?;
+        gas_wallets::table
+            .filter(gas_wallets::active.eq(false))
+            .first::<GasWallet>(&mut conn)
+            .await
+            .map_err(err_str!(FundsManagerError::Db))
+    }
+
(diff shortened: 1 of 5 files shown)


funds-manager/funds-manager-server/src/custody_client/queries.rs
```diff
@@ -27,6 +27,33 @@ impl CustodyClient {
         Ok(())
     }
 
+    /// Find an inactive gas wallet
+    pub async fn find_inactive_gas_wallet(&self) -> Result<GasWallet, FundsManagerError> {
+        let mut conn = self.get_db_conn().await?;
+        gas_wallets::table
+            .filter(gas_wallets::active.eq(false))
```
akirillo, line 34: clippy complaining here that there is no member `active` in the `gas_wallets` module - not sure what the diesel macros expand to but `active` doesn't seem to be a column on the `gas_wallets` table anyway, this should be `status` right?

akirillo (approved):
lgtm except for errors w/ the `find_inactive` / `mark_active` queries, but those seem like quick fixes so approving

funds-manager/funds-manager-server/src/custody_client/queries.rs
joeykraut: yeah this got messed up in the rebase order, it's fixed in one of the follow ups

PR #30, opened 2024-08-02
funds-manager: Refill gas to active wallets

joeykraut:
### Purpose
This PR refactors the way in which we refill gas in the `funds-manager`. Specifically, we now refill gas by looking up active gas wallets and topping off their balances directly.

### Testing
- Registered a new wallet as active then refilled its gas to 0.1 ETH

### DIFF
--- a/funds-manager/funds-manager-server/src/custody_client/gas_wallets.rs
+++ b/funds-manager/funds-manager-server/src/custody_client/gas_wallets.rs
@@ -10,6 +10,7 @@ use rand::thread_rng;
 use tracing::info;
 
 use crate::{
+    custody_client::DepositWithdrawSource,
     db::models::GasWalletStatus,
     error::FundsManagerError,
     helpers::{create_secrets_manager_entry_with_description, get_secret},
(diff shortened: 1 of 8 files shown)


funds-manager/funds-manager-server/src/custody_client/gas_wallets.rs
```diff
@@ -10,18 +11,52 @@ use rand::thread_rng;
...
+        fill_to: f64,
+    ) -> Result<(), FundsManagerError> {
+        // Fetch all active gas wallets
+        let active_wallets = self.get_active_gas_wallets().await?;
+
+        // Filter out those that don't need refilling
+        let mut wallets_to_fill: Vec<(String, f64)> = Vec::new(); // (address, fill amount)
+        for wallet in active_wallets {
+            let bal = self.get_ether_balance(&wallet.address).await?;
+            if bal + GAS_REFILL_TOLERANCE < fill_to || bal.is_zero() {
```
akirillo, line 47: we only need the `bal.is_zero()` check in the case that `GAS_REFILL_TOLERANCE >= fill_to`, which isn't something we'd ever expect, so perhaps we can remove the `bal.is_zero()` check?

funds-manager/funds-manager-server/src/custody_client/gas_wallets.rs
```diff
@@ -59,8 +94,9 @@ impl CustodyClient {
         let secret_name = Self::gas_wallet_secret_name(&gas_wallet.address);
         let secret_value = get_secret(&secret_name, &self.aws_config).await?;
 
-        // Update the gas wallet to be active and return the keypair
+        // Update the gas wallet to be active, top up wallets, and return the key
         self.mark_gas_wallet_active(&gas_wallet.address, peer_id).await?;
+        self.refill_gas_for_active_wallets(DEFAULT_TOP_UP_AMOUNT).await?;
```
akirillo, line 99: why do we top up all the active gas wallets here, as opposed to just the new one?

akirillo (changes requested):
this implementation looks great for relayer gas wallets, but we also need to be able to fund quoter wallets with gas. this is less frequent, as gas is only needed for permit2 approvals and transfers from quoter wallets -> hot wallet, but still necessary.

we could previously `withdraw-gas` directly to the quoter wallet, but now we would have to register an active gas wallet for each quoter wallet (with what peer ID?) and somehow prevent these wallets from being marked inactive when the `report-active-peers` endpoint is hit.

we should either have a separate endpoint for funding a quoter wallet w/ gas, or go through the same gas wallet registration flow for quoter gas wallets but somehow prevent them from being marked inactive.

funds-manager/funds-manager-server/src/custody_client/gas_wallets.rs
joeykraut: simpler, i see no reason not to

akirillo (approved):

PR #33, opened 2024-08-08
funds-manager: execution-client: Add quote execution helpers

joeykraut:
### Purpose
This PR sets up the execution flow for the `ExecutionClient` which involves:
1. Approving an ERC20 allowance for the 0x exchange proxy
2. Sending the transaction detailed by the quote received

### Testing
- Tested buying $GMX with $USDC on arbitrum using my personal wallet

### DIFF
--- a/funds-manager/funds-manager-server/src/execution_client/mod.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/mod.rs
@@ -2,9 +2,15 @@
 //! API
 pub mod error;
 pub mod quotes;
+pub mod swap;
 
 use std::sync::Arc;
 
+use ethers::{
+    middleware::SignerMiddleware,
+    providers::{Http, Provider},
+    signers::LocalWallet,
+};
 use reqwest::{Client, Url};
 use serde::Deserialize;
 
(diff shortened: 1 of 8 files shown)


funds-manager/funds-manager-server/src/execution_client/mod.rs
```diff
@@ -61,4 +80,12 @@ impl ExecutionClient {
             .await
             .map_err(ExecutionClientError::http)
     }
+
+    /// Get an instance of a signer middleware with the http provider attached
+    fn get_signer(
+        &self,
+        wallet: LocalWallet,
+    ) -> SignerMiddleware<Arc<Provider<Http>>, LocalWallet> {
+        SignerMiddleware::new(self.rpc_provider.clone(), wallet)
```
akirillo, line 89: we could attach the chain ID here for convenience

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.approve_erc20_allowance(
+            quote.sell_token_address,
+            quote.to,
+            quote.sell_amount,
+            &wallet,
+        )
+        .await?;
+
+        // Execute the swap
+        let receipt = self.execute_swap_tx(quote, wallet).await?;
```
akirillo, line 33: could we revoke the approved allowance if the swap execution fails? would prefer to leave no unused allowance on any external contracts

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -0,0 +1,80 @@
...
+    async fn execute_swap_tx(
+        &self,
+        quote: ExecutionQuote,
+        wallet: LocalWallet,
+    ) -> Result<TransactionReceipt, ExecutionClientError> {
+        let client = self.get_signer(wallet.clone());
+        let tx = Eip1559TransactionRequest::new()
+            .to(quote.to)
+            .from(quote.from)
+            .value(quote.value)
```
akirillo, line 67: we expect this to always be 0, right? might as well exclude it here (and maybe from the `ExecutionQuote` struct), just to bar against any bugs in the 0x api

akirillo (approved):

funds-manager/funds-manager-server/src/execution_client/swap.rs
joeykraut: you'll see in a follow up that is not possible as we separate out allowances from execution.

funds-manager/funds-manager-server/src/execution_client/swap.rs
joeykraut: their API does not guarantee this to be zero, i'll go with what they've claimed in their docs

funds-manager/funds-manager-server/src/execution_client/mod.rs
joeykraut: this is done in a follow up on the wallet directly

PR #35, opened 2024-08-10
funds-manager: handlers: Handle quote fetch and execution

joeykraut:
### Purpose
This PR implements the quote fetch and execution endpoints for quoter rebalancing. In addition to the straightforward calls to the `ExecutionClient`, I also moved ERC20 approvals to the quote phase. This is necessary as the 0x API will not provide a quote unless the sell side balance is already approved.

### Testing
- Tested on Arbitrum mainnet by funding the quoter wallet on mainnet from my personal wallet. Hit the endpoint to fetch and execute on a quote to buy $GMX.

### DIFF
--- a/funds-manager/funds-manager-server/src/execution_client/quotes.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/quotes.rs
@@ -1,7 +1,16 @@
 //! Client methods for fetching quotes and prices from the execution venue
 
+use std::{str::FromStr, sync::Arc};
+
+use ethers::{
+    signers::{LocalWallet, Signer},
+    types::{Address, U256},
+};
 use funds_manager_api::quoters::ExecutionQuote;
 use serde::Deserialize;
+use tracing::info;
+
+use crate::helpers::ERC20;
 
 use super::{error::ExecutionClientError, ExecutionClient};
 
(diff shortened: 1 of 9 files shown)


funds-manager/funds-manager-server/src/execution_client/quotes.rs
```diff
@@ -45,19 +62,60 @@ impl ExecutionClient {
...
+        buy_asset: Address,
+        sell_asset: Address,
         amount: u128,
-        recipient: &str,
+        wallet: &LocalWallet,
     ) -> Result<ExecutionQuote, ExecutionClientError> {
+        // First, set an approval for the sell token, the 0x api will not give a quote
+        // if its contract is not an approved spender for the requested amount
+        let exchange_addr = Address::from_str(EXCHANGE_PROXY_ADDRESS).unwrap();
+        self.approve_erc20_allowance(sell_asset, exchange_addr, U256::from(amount), wallet).await?;
```
akirillo, line 73: my preference is still to err on the side of not having unused approvals on external contracts, which would imply setting the allowance back down to 0 after hitting the quote endpoint, and invoking another approval ahead of executing a swap (w/ revocation in the case of a failed swap). I think this is worth the gas cost & complexity, just another approval call. lmk what u think

akirillo (approved):

funds-manager/funds-manager-server/src/execution_client/quotes.rs
joeykraut: let's see how this goes and decide whether we want that. i really don't find covering all possible dangers on any margin of implied complexity to be a productive path.

PR #37, opened 2024-08-24
funds-manager: Refill all gas wallets, not just active

joeykraut:
### Purpose
This PR makes a few miscellaneous changes to the funds manager:
- Adds logging for transitions in gas wallet states
- Refills _all_ gas wallets at a time, as opposed to just active ones. This adds robustness in the case of an incorrectly expired wallet and costs us no more -- unused wallets will be full anyways.
- Changes the logic for finding an asset id by its erc20 mint to be case insensitive


funds-manager/funds-manager-server/src/custody_client/gas_wallets.rs
```diff
@@ -31,13 +31,13 @@ impl CustodyClient {
     // | Handlers |
     // ------------
 
-    /// Refill gas for all active wallets
+    /// Refill gas for all gas wallets
     pub(crate) async fn refill_gas_for_active_wallets(
```
akirillo, line 35: nit: could the function name reflect this change as well, i.e. `refill_all_gas_wallets`

also the comment below above the `get_all_gas_wallets` call

akirillo (approved):

PR #41, opened 2024-08-30
funds-manager-server: set max allowable withdrawal value

akirillo:
This PR enforces an upper bound the amount that can be withdrawn from the funds manager. This is to protect against any unforeseen errant behavior from the quoters. This value is currently hardcoded to $50k.

### DIFF
--- a/funds-manager/funds-manager-server/src/handlers.rs
+++ b/funds-manager/funds-manager-server/src/handlers.rs
@@ -21,6 +21,7 @@ use itertools::Itertools;
 use serde_json::json;
 use std::collections::HashMap;
 use std::sync::Arc;
+use tracing::warn;
 use warp::reply::Json;
 
 /// The "mints" query param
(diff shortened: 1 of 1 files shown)


funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -89,6 +92,27 @@ pub(crate) async fn quoter_withdraw_handler(
...
     server: Arc<Server>,
 ) -> Result<Json, warp::Rejection> {
+    // Get the price of the token
+    let maybe_price = server
+        .relayer_client
+        .get_binance_price(&withdraw_request.mint)
+        .await
+        .map_err(|e| warp::reject::custom(ApiError::InternalError(e.to_string())))?;
+
+    if let Some(price) = maybe_price {
```
joeykraut, line 102: can we log a warning or error here if there's no price? want to make sure that this doesn't go unnoticed

joeykraut (approved):

PR #42, opened 2024-09-04
price-reporter: Make misc reliability improvements

joeykraut:
### Purpose
This PR makes a few changes to the price reporter:
- Changes the underlying synchronization primitive to a `Watch` which 1. streams immediately, and 2. more directly captures the semantics of the price reporter.
- Using a watch means that a websocket and http request both _instantly_ get a price
- Initialize default pair streams on startup
- Do not remove a stream once clients disconnect

### Testing
- Tested all websocket pairs, verified that they began streaming instantly

### DIFF
--- a/price-reporter/src/main.rs
+++ b/price-reporter/src/main.rs
@@ -11,7 +11,12 @@ use std::net::SocketAddr;
 
 use errors::ServerError;
 use http_server::HttpServer;
+use renegade_common::types::{
+    exchange::Exchange,
+    token::{Token, TOKEN_REMAPS, USDC_TICKER, USDT_TICKER, USD_TICKER},
+};
 use renegade_config::setup_token_remaps;
+use renegade_price_reporter::{manager::get_listing_exchanges, worker::ExchangeConnectionsConfig};
 use renegade_util::err_str;
 use tokio::{net::TcpListener, sync::mpsc::unbounded_channel};
 use tracing::{error, info};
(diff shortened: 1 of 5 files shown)


price-reporter/src/main.rs
```diff
@@ -80,3 +89,75 @@ async fn main() -> Result<(), ServerError> {
...
+    config: ExchangeConnectionsConfig,
+) -> Result<(), ServerError> {
+    info!("Initializing default price streams");
+
+    // Get the default token remap
+    let quote_token = Token::from_ticker(DEFAULT_STABLE);
+    let remap = TOKEN_REMAPS.get().unwrap();
+    for (addr, ticker) in remap.clone().into_iter() {
+        // Skip stables
+        if [USDT_TICKER, USDC_TICKER, USDT_TICKER].contains(&ticker.as_str()) {
```
akirillo, line 105: nit: listed `USDT_TICKER` twice

price-reporter/src/main.rs
```diff
@@ -80,3 +89,75 @@ async fn main() -> Result<(), ServerError> {
...
+            )
+            .await?;
+        }
+    }
+
+    Ok(())
+}
+
+/// Spawn a task to initialize a price stream for a given token pair
+async fn init_price_stream(
```
akirillo, line 127: clippy calling out unnecessary `async` here

akirillo (approved):

PR #43, opened 2024-09-06
price-reporter: ws-server: Retry connection on send failure

joeykraut:
### Purpose
This PR makes two changes:
1. We refactor the `price_stream_task` into a retry wrapper that calls `manage_connection`. This captures a broader set of errors that may be retried, including error sending a keepalive on a closed connection. Looking at our logs, this makes up the majority of server failures
2. Adds logging for the case in which re-establishing a connection fails

### Testing
- Ran the price reporter locally and simulated a failure, verified that the recovery logic worked as expected


akirillo (approved):
clean

PR #45, opened 2024-09-12
events-collector: events-collector-server: create unaugmented event schema & models

akirillo:
This PR initializes the `events-collector-server` crate w/ `diesel` definitions of the DB schema & models, and an initial migration (which was applied to a dev instance of the DB).

For now, our schema simply mirrors the fundamental event types, augmented only with a unique ID and timestamp for each event.

In the future, this schema will diverge further from the underlying events as we add additional "augmentations" such as price at time of event, time since last fill, etc.

TODO:
- Basic server configuration
- Basic augmentation framework


events-collector/events-collector-server/src/db/models.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub event_id: Uuid,
+    /// The event timestamp
+    pub event_timestamp: SystemTime,
+
+    /// The ID of the wallet that made the deposit
+    pub wallet_id: Uuid,
+    /// The mint of the asset that was deposited
+    pub mint: String,
+    /// The amount of the deposit
+    pub amount: BigDecimal,
```
joeykraut, line 38: can we call out here whether this is decimal adjusted or not? and same for other fields that specify an erc20 amount

events-collector/migrations/2024-09-12-002231_create_initial_event_tables/down.sql
```diff
@@ -0,0 +1,9 @@
+-- This file should undo anything in `up.sql`
```
joeykraut, line 1: remove the autogen comment

events-collector/migrations/2024-09-12-002231_create_initial_event_tables/up.sql
```diff
@@ -0,0 +1,76 @@
+-- Your SQL goes here
```
joeykraut, line 1: same here, please add a better comment in place

joeykraut (approved):

PR #50, opened 2024-10-21
auth: auth-server: Define key management API

joeykraut:
### Purpose
This PR defines a minimal key management API; specifically with an endpoint to add a key and another to deprecate a key.

We store API secrets encrypted under the server's symmetric encryption key, this PR adds that logic as well.

### Testing
- Tested both new endpoints, verified the database state after testing


auth/auth-server-api/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+// ----------------------
+
+/// The path to create a new API key
+///
+/// POST /api-keys
+pub const API_KEYS_PATH: &str = "api-keys";
+/// The path to mark an API key as inactive
+///
+/// POST /api-keys/{id}/deactivate
+pub const DEACTIVATE_API_KEY_PATH: &str = "deactivate";
```
akirillo, line 23: why not make this the entire templated path, i.e. `api-keys/:id/deactivate`? it's confusing in client-side code to import this constant and still have to prepend the first path segments

auth/auth-server-api/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Debug, Deserialize)]
+pub struct CreateApiKeyRequest {
+    /// The API key id
+    pub id: Uuid,
+    /// The API key secret
+    ///
+    /// Expected as a base64 encoded string
+    pub secret: String,
+    /// The name of the API key
+    pub name: String,
```
akirillo, line 35: nit: can we either rename this to `username`, or rename the column on the `ApiKey` table to `name`, so that we refer to these fields consistently when interfacing w/ the API and the DB?

auth/auth-server-api/src/lib.rs
akirillo: ah, I see how this is useful in `auth-server/main.rs` when using the `warp::path::param`. still, for client-side ergonomics, it may be a good idea to export a separate constant / method that constructs the full / templated path

auth/auth-server/src/main.rs
```diff
@@ -82,19 +90,43 @@ async fn main() {
...
+        .and(with_server(server.clone()))
+        .and_then(|request, server: Arc<Server>| async move { server.add_key(request).await });
+
+    // Expire an API key
+    let expire_api_key = warp::path(API_KEYS_PATH)
+        .and(warp::path::param::<Uuid>())
+        .and(warp::path(DEACTIVATE_API_KEY_PATH))
+        .and(warp::post())
+        .and(with_server(server.clone()))
+        .and_then(|id: Uuid, server: Arc<Server>| async move { server.expire_key(id).await });
```
akirillo, line 128: what's our plan for authenticating these endpoints? introduce another symmetric key and use HMAC auth like we do elsewhere? or am I mistaken in thinking we'll need these endpoints authenticated?

akirillo (approved):

auth/auth-server-api/src/lib.rs
joeykraut: it's not quite a username, description is most appropriate, i can change it to that

auth/auth-server/src/main.rs
joeykraut: that's correct, we'll have an API key for admin operations here.

PR #58, opened 2024-11-22
price-reporter: use new token mapping, use default exchange stable

akirillo:
This PR updates the price reporter extension to use the new token mapping logic in the relayer (by way of upgrading the relayer crate versions), and tweaks the logic in initializing the default price streams to use the default stable for each exchange as the quote token.

Additionally, we fix some compilation errors around creating new wallets in the funds manager as a result of the version bump.

I test the price reporter by running it locally, using the testnet token mapping, and ensuring that each price stream was initialized without error. I also tested adding a new token ([MAGIC](https://app.uniswap.org/explore/tokens/arbitrum/0x539bdE0d7Dbd336b79148AA742883198BBF60342)) to a local token mapping file, and asserted that all supported exchanges stream it out-of-the box


Cargo.toml
```diff
@@ -22,18 +22,18 @@ lto = true
...
-renegade-price-reporter = { package = "price-reporter", git = "https://github.com/renegade-fi/renegade.git" }
+renegade-api = { package = "external-api", git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
+renegade-common = { package = "common", git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
+renegade-config = { package = "config", git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
+renegade-constants = { package = "constants", git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
+renegade-circuits = { package = "circuits", git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
+renegade-circuit-types = { package = "circuit-types", git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
+renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
+renegade-util = { package = "util", git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/new-token-mapping" }
```
joeykraut, line 35: changes like this probably don't need to be committed in the future

price-reporter/src/main.rs
```diff
@@ -107,11 +103,14 @@ async fn init_default_price_streams(
         }
 
         let base_token = Token::from_addr(&addr);
-        let supported_exchanges = get_supported_exchanges(&base_token, &quote_token, &config);
+        let supported_exchanges = get_supported_exchanges(&base_token, &config);
```
joeykraut, line 106: from clippy: this method doesn't need to be async

joeykraut (approved):

PR #61, opened 2024-12-06
price-reporter: routes: add admin token mapping refresh route

akirillo:
This PR adds an authenticated route to the price reporter for refreshing the token mapping. The authentication scheme is inherited from the relayer, and in accordance with this the price reporter now accepts an HMAC key in its configuration.

After the token mapping is refreshed, price streams are reinitialized. This is a no-op for existing price streams but will start streams for any newly-added tokens. Note that in the case of tokens being _removed_ from the mapping, the price streams will not be closed - but practically this is not a concern.

This was tested locally, with authentication disabled and using a local token mapping file instead of modifying the actual repo. This will be tested end-to-end in testnet shortly.

### DIFF
--- a/price-reporter/src/http_server/routes.rs
+++ b/price-reporter/src/http_server/routes.rs
@@ -1,12 +1,17 @@
 //! The routes for the HTTP server
 
 use async_trait::async_trait;
-use hyper::{Body, Request, Response, StatusCode};
-use renegade_common::types::Price;
+use hyper::{body::to_bytes, Body, Request, Response, StatusCode};
+use renegade_api::auth::validate_expiring_auth;
+use renegade_arbitrum_client::constants::Chain;
+use renegade_common::types::{wallet::keychain::HmacKey, Price};
+use renegade_config::setup_token_remaps;
 use renegade_price_reporter::worker::ExchangeConnectionsConfig;
+use renegade_util::err_str;
 
 use crate::{
     errors::ServerError,
+    init_default_price_streams,
     utils::{parse_pair_info_from_topic, UrlParams},
     ws_server::GlobalPriceStreams,
 };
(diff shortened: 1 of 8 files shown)


price-reporter/src/http_server/routes.rs
```diff
@@ -98,3 +103,83 @@ impl Handler for PriceHandler {
...
+    async fn handle(&self, mut req: Request<Body>, _: UrlParams) -> Response<Body> {
+        match self.authenticate_request(&mut req).await {
+            Ok(_) => (),
+            Err(e) => {
+                return Response::builder()
+                    .status(StatusCode::UNAUTHORIZED)
+                    .body(Body::from(e.to_string()))
+                    .unwrap()
+            },
+        }
```
joeykraut, line 175: Better to use `if let` here if we only handle one branch

price-reporter/src/http_server/routes.rs
```diff
@@ -98,3 +103,83 @@ impl Handler for PriceHandler {
...
+    pub async fn refresh_token_mapping(&self) -> Result<(), ServerError> {
+        let token_remap_path = self.token_remap_path.clone();
+        let remap_chain = self.remap_chain;
+        tokio::task::spawn_blocking(move || setup_token_remaps(token_remap_path, remap_chain))
+            .await
+            .map_err(err_str!(ServerError::TokenRemap))
+            .and_then(|res| res.map_err(err_str!(ServerError::TokenRemap)))?;
+
+        // Re-initialize the default price streams after refreshing the token mapping
+        init_default_price_streams(&self.price_streams, &self.config)
```
joeykraut, line 160: qq for my own understanding: what will happen to the existing price streams? are they shut down?

price-reporter/src/utils.rs
```diff
@@ -128,6 +131,8 @@ pub struct PriceReporterConfig {
     pub remap_chain: Chain,
     /// The configuration options that may be used by exchange connections
     pub exchange_conn_config: ExchangeConnectionsConfig,
+    /// The HMAC key for the admin API
+    pub admin_key: HmacKey,
```
joeykraut, line 135: can we make this optional and disable the refresh endpoint if that's the case? helpful for testing locally

joeykraut (approved):

price-reporter/src/http_server/routes.rs
akirillo: existing price streams are untouched, we only initialize new price streams if they are not already present in the global mapping

PR #62, opened 2024-12-10
auth server: emit metrics from atomic match path

sehyunc:
### Purpose
This PR adds functionality to record metrics when proxying external match requests. We calculate the quote per base from the match bundle in the response. Recorded metrics include:
- base / quote amounts for external orders
- base / quote amounts for external matches

Note that this PR does not include recording settlement events, that will be in a separate PR.

### Testing
Tested with a locally running auth server and Datadog agent.
Need to test in testnet environment.


auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -66,15 +66,22 @@ impl Server {
...
         // Log the bundle parameters
         if let Err(e) = self.log_bundle(resp.body()) {
             warn!("Error logging bundle: {e}");
         }
+
+        // Record metrics
+        if let Err(e) = record_external_match_metrics(&body, resp.body(), key_description).await {
+            warn!("Error recording metrics: {e}");
+        }
+
```
sehyunc, line 84: Should we spawn metric recording into the background using Tokio to prevent added latency?

auth/auth-server/src/error.rs
```diff
@@ -25,6 +25,9 @@ pub enum AuthServerError {
     /// Error setting up the auth server
     #[error("Error setting up the auth server: {0}")]
     Setup(String),
+    /// An error setting up the token remapping
+    #[error("Error setting up the token remapping: {0}")]
+    TokenRemap(String),
```
joeykraut, line 30: probably makes more sense to just use the `Setup` error for this, no need to define a new enum variant

auth/auth-server/src/server/api_auth.rs
```diff
@@ -40,7 +40,7 @@ impl Server {
 
         let key_description = self.check_api_key_auth(api_key, path, headers, body).await?;
         info!("Authorized request for entity: {key_description}");
-        Ok(())
+        Ok(key_description)
```
joeykraut, line 43: nit: Let's add another line to the doc string for this method indicating that it returns a human readable description of the key

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,154 @@
...
+    let labels = vec![
+        (ASSET_METRIC_TAG.to_string(), asset),
+        (KEY_DESCRIPTION_METRIC_TAG.to_string(), key_description),
+    ];
+
+    metrics::counter!(metric_name, &labels).increment(1);
+}
+
+/// Records all metrics related to an external match request and response
+pub async fn record_external_match_metrics(
```
joeykraut, line 123: don't see any `await`s in this method, so it doesn't need to be async. Will you add `#![deny(clippy::unused_async)]` to the top of `main.rs` so that clippy will catch this in the future?

auth/auth-server/src/server/handle_external_match.rs
joeykraut: yeah probably best, just so we don't have to worry about it, you'll want `spawn_blocking` because it looks like all the metrics recorder implementations use a blocking IPC

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,154 @@
...
+    let token = Token::from_addr(mint);
+    let asset = token.get_ticker().unwrap_or(mint.to_string());
+    let volume = token.convert_to_decimal(amount);
+
+    (asset, volume)
+}
+
+/// Calculates the quote per base price from a match bundle
+/// Returns the price as an f64 decimal adjusted value
+fn calculate_quote_per_base_price(match_bundle: &AtomicMatchApiBundle) -> f64 {
```
joeykraut, line 35: nit: `calculate_implied_price` is probably a better name here

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,154 @@
...
+    let (quote, base) = match match_bundle.match_result.direction {
+        OrderSide::Buy => (&match_bundle.send, &match_bundle.receive),
+        OrderSide::Sell => (&match_bundle.receive, &match_bundle.send),
+    };
+
+    Token::from_addr(&quote.mint).convert_to_decimal(quote.amount)
+        / Token::from_addr(&base.mint).convert_to_decimal(base.amount)
```
joeykraut, line 42: nit: I'd break this up into a few smaller lines:
```rust
let quote_amt = Token::from_addr(&quote.mint).convert_to_decimal(quote.amount);
let base_amt = Token::from_addr(&base.mint).convert_to_decimal(base.amount);
quote_amt / base_amt
```
Slightly more readable to an unfamiliar reader in that the intermediate variable names help annotate the code

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/// Converts a decimal amount to token native units, accounting for the token's
+/// decimals. This is the inverse operation of convert_to_decimal.
+fn convert_from_decimal(token: &Token, decimal_amount: f64) -> u128 {
+    let decimals = token.get_decimals().unwrap_or_default();
+    (decimal_amount * (10u128.pow(decimals as u32) as f64)) as u128
```
joeykraut, line 49: same here as above, smaller steps would make reading this trivial:
```rust
let decimal_correction = 10f64.powi(decimals as i32);
let corrected_amount = decimal_amount * decimal_correction;
corrected_amount as u128
```

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,154 @@
...
+    price: f64,
+    labels: &[(String, String)],
+) -> Result<(), AuthServerError> {
+    // Record external order volume
+    let base_mint = biguint_to_hex_addr(&req.external_order.base_mint);
+    let quote_mint = biguint_to_hex_addr(&req.external_order.quote_mint);
+
+    // Calculate amount in base
+    let timestamped_price = TimestampedPrice::new(price);
+    let fixed_point_price = timestamped_price.as_fixed_point();
```
joeykraut, line 81: you can convert from `f64 -> FixedPoint` more directly using `FixedPoint::from_f64_round_down`

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,154 @@
...
+    let order = req.external_order.to_order_with_price(fixed_point_price);
+
+    // Calculate amount in quote
+    let (_, volume) = get_asset_and_volume(&base_mint, order.amount);
+    let quote_token = Token::from_addr(&quote_mint);
+    let quote_amount = convert_from_decimal(&quote_token, volume * fixed_point_price.to_f64());
```
joeykraut, line 87: no need to convert back the `fixed_point_price` price to `f64`, just use `price`.

auth/auth-server/src/telemetry/labels.rs
```diff
@@ -0,0 +1,28 @@
+//! Metric names and tags
+
+// ----------------
+// | METRIC NAMES |
+// ----------------
+
+/// Metric describing the number of atomic matches requested
+pub const NUM_ATOMIC_MATCH_REQUESTS: &str = "num_atomic_match_requests";
```
joeykraut, line 8: nit: `NUM_EXTERNAL_MATCH_REQUESTS` is a better name for consistency

joeykraut (approved):
nice, some room for improvement, but just small things

PR #63, opened 2024-12-11
auth server: record settlement metrics

sehyunc:
### Purpose
This PR records metrics on external match settlement. 

### Testing
Tested locally against locally running auth server and Datadog agent. Example dashboard from this data [here](https://us5.datadoghq.com/dashboard/6j9-i8d-xdq?fromUser=true&refresh_mode=paused&from_ts=1733870660790&to_ts=1733899576032&live=false).
Need to test in testnet stack.

### DIFF
--- a/auth/auth-server/src/main.rs
+++ b/auth/auth-server/src/main.rs
@@ -22,13 +22,18 @@ mod telemetry;
 
 use auth_server_api::API_KEYS_PATH;
 use clap::Parser;
-use renegade_arbitrum_client::constants::Chain;
+use ethers::signers::LocalWallet;
+use renegade_arbitrum_client::{
+    client::{ArbitrumClient, ArbitrumClientConfig},
+    constants::Chain,
+};
 use renegade_config::setup_token_remaps;
 use renegade_util::err_str;
 use renegade_util::telemetry::configure_telemetry;
 use reqwest::StatusCode;
 use serde_json::json;
 use std::net::SocketAddr;
+use std::str::FromStr;
 use std::sync::Arc;
 use thiserror::Error;
 use tracing::{error, info};
@@ -39,6 +44,12 @@ use server::Server;
 
 /// The default internal server error message
 const DEFAULT_INTERNAL_SERVER_ERROR_MESSAGE: &str = "Internal Server Error";
+/// The dummy private key used to instantiate the arbitrum client
+///
+/// We don't need any client functionality using a real private key, so instead
+/// we use the key deployed by Arbitrum on local devnets
+const DUMMY_PRIVATE_KEY: &str =
+    "0xb6b15c8cb491557369f3c7d2c287b053eb229daa9c22138887752191c9520659";
 
 // -------
 // | CLI |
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/main.rs
```diff
@@ -163,8 +174,22 @@ async fn main() {
...
+            .to_string(),
+        arb_priv_keys: vec![wallet],
+        block_polling_interval_ms: 100,
+    })
+    .await
+    .unwrap();
+
     // Create the server
-    let server = Server::new(args).await.expect("Failed to create server");
+    let server = Server::new(args, arbitrum_client).await.expect("Failed to create server");
```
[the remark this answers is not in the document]
sehyunc: Moving forward, perhaps a `TelemetryClient` entity would be a better place to instantiate the Arbitrum client since we'll be adding more complex logic regarding quote comparison and on-chain queries

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -77,10 +77,14 @@ impl Server {
             warn!("Error logging bundle: {e}");
         }
 
-        // Record metrics in a blocking task
+        // Record metrics
         let resp_clone = resp.body().to_vec();
-        tokio::task::spawn_blocking(move || {
-            if let Err(e) = record_external_match_metrics(&body, &resp_clone, key_description) {
+        let arbitrum_client = self.arbitrum_client.clone();
```
joeykraut, line 82: will you also move the logging lines above into this async closure?

auth/auth-server/src/main.rs
```diff
@@ -39,6 +44,12 @@ use server::Server;
 
 /// The default internal server error message
 const DEFAULT_INTERNAL_SERVER_ERROR_MESSAGE: &str = "Internal Server Error";
+/// The dummy private key used to instantiate the arbitrum client
+///
+/// We don't need any client functionality using a real private key, so instead
+/// we use the key deployed by Arbitrum on local devnets
+const DUMMY_PRIVATE_KEY: &str =
```
joeykraut, line 51: i'll add this to my todo to remove the need for a private key. then we can take the dummy values out of our workers

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -148,9 +201,66 @@ pub fn record_external_match_metrics(
...
+async fn await_settlement(
+    match_bundle: &AtomicMatchApiBundle,
+    arbitrum_client: &ArbitrumClient,
+) -> Result<bool, AuthServerError> {
+    let nullifier = extract_nullifier_from_match_bundle(match_bundle)?;
+    let res = arbitrum_client.await_nullifier_spent(nullifier, ATOMIC_SETTLEMENT_TIMEOUT).await;
+
+    let did_settle = res.is_ok();
+    if !did_settle {
+        warn!("atomic match settlement not observed on-chain");
```
joeykraut, line 231: can we change this to `info`? this is the common case at the moment and it doesn't directly indicate that anything is wrong. been meaning to do that in the relayer

joeykraut (approved):

PR #64, opened 2024-12-11
auth server: add request-level tag

sehyunc:
### Purpose
This PR adds metric tags to allow for richer queries in DD. Specifically, we want to 
- correlate different components of a request together (order, bundle, settlement)
- indicate base asset on quote volume metrics to group by base asset

### Testing
Tested against locally running auth server and Datadog agent.

### DIFF
--- a/auth/auth-server/src/telemetry/helpers.rs
+++ b/auth/auth-server/src/telemetry/helpers.rs
@@ -23,10 +23,11 @@ use tracing::{info, warn};
 use crate::{
     error::AuthServerError,
     telemetry::labels::{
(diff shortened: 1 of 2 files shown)


auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -69,6 +70,20 @@ fn convert_from_decimal(token: &Token, decimal_amount: f64) -> u128 {
     corrected_amount as u128
 }
 
+/// Extends the given labels with a base asset tag
+fn extend_labels_with_base_asset(
+    base_mint: &str,
+    labels: &[(String, String)],
```
joeykraut, line 76: you could take a mutable reference here, not that opinionated though

joeykraut (approved):

PR #74, opened 2024-12-27
auth-server: record fill ratio metrics

sehyunc:
### Purpose
This PR records fill ratios of quote requests and external match requests (to be deprecated) for display as histograms on a Datadog dashboard. Also records number of quote requests per key.

### Testing
- [x] Tested locally
- [ ] Tested in testnet

### DIFF
--- a/auth/auth-server/src/telemetry/helpers.rs
+++ b/auth/auth-server/src/telemetry/helpers.rs
@@ -23,7 +23,7 @@ use crate::{
     error::AuthServerError,
     telemetry::labels::{
         ASSET_METRIC_TAG, BASE_ASSET_METRIC_TAG, DECIMAL_CORRECTION_FIXED_METRIC_TAG,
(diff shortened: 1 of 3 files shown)


auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -225,6 +236,16 @@ pub(crate) async fn record_external_match_metrics(
         warn!("Error recording request metrics: {e}");
     }
 
+    // Record fill ratio metric
+    let requested_quote_amount = order.get_quote_amount(FixedPoint::from_f64_round_down(price));
+    let matched_quote = match match_resp.match_bundle.match_result.direction {
+        OrderSide::Buy => &match_resp.match_bundle.send,
+        OrderSide::Sell => &match_resp.match_bundle.receive,
+    };
```
joeykraut, line 244: same thing here, can just directly use the `quote_amount` field on the match result

joeykraut (approved):

PR #75, opened 2024-12-28
auth-server: server: rate-limiter: Implement bundle rate limits

joeykraut:
### Purpose
This PR implements bundle rate limits for the auth server. Specifically, we rate limit the number of inflight, unsettled bundles per client. So, a client may accrue new API tokens by either:
- Waiting for their token bucket to refill
- Settling a match

### Testing
- [ ] Test in testnet

### DIFF
--- a/auth/auth-server/src/server/handle_external_match.rs
+++ b/auth/auth-server/src/server/handle_external_match.rs
@@ -5,13 +5,16 @@
 
 use bytes::Bytes;
 use http::Method;
-use renegade_api::http::external_match::{ExternalMatchResponse, ExternalQuoteResponse};
+use renegade_api::http::external_match::{
+    AssembleExternalMatchRequest, ExternalMatchRequest, ExternalMatchResponse, ExternalOrder,
+    ExternalQuoteResponse,
+};
 use tracing::{info, instrument, warn};
 use warp::{reject::Rejection, reply::Reply};
 
 use super::Server;
 use crate::error::AuthServerError;
-use crate::telemetry::helpers::record_external_match_metrics;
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/server/rate_limiter.rs
```diff
@@ -0,0 +1,70 @@
+//! A rate limiter for the server
+//!
+//! The unit which we rate limit is number of inflight bundles. Therefore, there
+//! are two ways for the token bucket to refill:
+//!     - Wait for the next refill
+//!     - Settle a bundle on-chain
+//!
+//! The ladder is measured by waiting for nullifier spend events on-chain
```
sehyunc, line 8: *latter

sehyunc (approved):
clean

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -67,36 +79,79 @@ impl Server {
...
+
+        // If the bundle settles, increase the API user's a rate limit token balance
+        let did_settle = await_settlement(&match_resp.match_bundle, &self.arbitrum_client).await?;
+        if did_settle {
+            self.add_rate_limit_token(key.clone()).await;
+        }
+
+        // Log the bundle and record metrics
+        self.log_bundle(resp)?;
+        record_external_match_metrics(&order, match_resp, key, did_settle).await
```
sehyunc, line 152: I'm assuming we're ok with using the same metric name for external match and quote and assemble endpoints?

joeykraut: yeah i think we care more about the aggregate

PR #82, opened 2025-01-11
auth-server: update quote comparison impl

sehyunc:
### Purpose
This PR refactors auth-server quote comparison metrics by:
- Moving metrics collection to bundle response for more realistic comparisons
- Parallelizing quote source fetching
- Implementing async QuoteSource trait in preparation for new sources

### Testing
- [x] Tested locally
- [ ] Tested in testnet

### DIFF
--- a/auth/auth-server/src/server/handle_external_match.rs
+++ b/auth/auth-server/src/server/handle_external_match.rs
@@ -163,9 +163,17 @@ impl Server {
             (REQUEST_ID_METRIC_TAG.to_string(), uuid::Uuid::new_v4().to_string()),
             (DECIMAL_CORRECTION_FIXED_METRIC_TAG.to_string(), "true".to_string()),
         ];
-        self.quote_metrics
-            .record_quote_comparison(&match_resp.match_bundle, labels.as_slice())
-            .await;
+
+        // Spawn quote comparisons and collect handles
+        let comparison_handles =
+            self.quote_metrics.record_quote_comparison(&match_resp.match_bundle, labels.as_slice());
+
+        // Handle quote comparison results
+        for handle in comparison_handles {
+            if let Err(e) = handle.await {
+                warn!("Quote comparison failed: {e}");
+            }
+        }
 
         // If the bundle settles, increase the API user's a rate limit token balance
         let did_settle = await_settlement(&match_resp.match_bundle, &self.arbitrum_client).await?;
@@ -175,7 +183,7 @@ impl Server {
 
         // Log the bundle and record metrics
         self.log_bundle(resp)?;
-        record_external_match_metrics(&order, match_resp.clone(), key, did_settle).await?;
+        record_external_match_metrics(&order, match_resp, key, did_settle).await?;
 
         Ok(())
     }
--- a/auth/auth-server/src/server/handle_external_match.rs
+++ b/auth/auth-server/src/server/handle_external_match.rs
@@ -47,7 +47,7 @@ impl Server {
         let resp_clone = resp.body().to_vec();
         let server_clone = self.clone();
         tokio::spawn(async move {
-            if let Err(e) = server_clone.handle_quote_response(key_desc, &body, &resp_clone) {
+            if let Err(e) = server_clone.handle_quote_response(key_desc, &body, &resp_clone).await {
                 warn!("Error handling quote: {e}");
             }
         });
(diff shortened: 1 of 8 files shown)


auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -165,7 +175,9 @@ impl Server {
 
         // Log the bundle and record metrics
         self.log_bundle(resp)?;
-        record_external_match_metrics(&order, match_resp, key, did_settle).await
+        record_external_match_metrics(&order, match_resp.clone(), key, did_settle).await?;
```
joeykraut, line 178: any reason we clone the match response here? doesn't look like it gets used

auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -260,7 +249,7 @@ pub(crate) async fn record_external_match_metrics(
     ];
 
     // Get decimal-corrected price
-    let price = calculate_implied_price(&match_resp.match_bundle)?;
+    let price = calculate_implied_price(&match_resp.match_bundle, true)?;
```
joeykraut, line 252: nit: it's good to annotate inlined literals that aren't self-describing:
```rust
let price = calculate_implied_price(&match_resp.match_bundle, true /* decimal_correct */)?;
```

auth/auth-server/src/telemetry/quote_comparison.rs
```diff
@@ -20,50 +21,67 @@ pub struct QuoteComparison {
...
+            let base_token = base_token.clone();
+            let quote_token = quote_token.clone();
+            let side = match_bundle.match_result.direction;
+            let source_name = source.name().to_string();
+
+            async move {
+                let quote =
+                    source.get_quote(base_token, quote_token, side, amount, our_price).await;
+                (source_name, quote)
+            }
```
joeykraut, line 70: you'll probably want to explicitly spawn this on a tokio thread with `tokio::spawn`

joeykraut (approved):

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -163,9 +163,17 @@ impl Server {
...
+        // Spawn quote comparisons and collect handles
+        let comparison_handles =
+            self.quote_metrics.record_quote_comparison(&match_resp.match_bundle, labels.as_slice());
+
+        // Handle quote comparison results
+        for handle in comparison_handles {
+            if let Err(e) = handle.await {
+                warn!("Quote comparison failed: {e}");
+            }
+        }
```
joeykraut, line 176: Couple things here:
1. I'd move the wait logic into the callee, so that `record_quote_comparison` cleanly encapsulates any concurrent behavior.
2. Better to use [`futures::join_all`](https://docs.rs/futures/latest/futures/future/fn.join_all.html) here. For Tokio in particular this is fine, but note that if you had a vector of futures in general, awaiting them in a loop will actually execute them one after another -- not concurrently. The way to think about asynchrony in Rust is "futures do nothing unless awaited"

auth/auth-server/src/telemetry/quote_comparison.rs
```diff
@@ -56,33 +57,30 @@ impl QuoteComparisonHandler {
...
         };
 
-        // Create all quote futures
-        let quote_futures = self.sources.iter().map(|source| {
-            let base_token = base_token.clone();
-            let quote_token = quote_token.clone();
-            let side = match_bundle.match_result.direction;
-            let source_name = source.name().to_string();
+        // Spawn parallel quote fetching and comparison tasks
+        self.sources
```
joeykraut, line 61: nit: i'd use a for loop over `self.sources` here. the iterator adds a lot of nested control flow and structure that hurts readability. long `map` methods are usually harder to read

auth/auth-server/src/telemetry/quote_comparison.rs
```diff
@@ -56,33 +57,30 @@ impl QuoteComparisonHandler {
...
-            record_comparison(&comparison, &labels);
-        }
+                    let price_diff_bips = calculate_price_diff_bps(our_price, quote.price, is_sell);
+                    let comparison = QuoteComparison {
+                        our_price,
+                        source_price: quote.price,
+                        source_name: source.name().to_string(),
+                        price_diff_bips,
+                    };
+                    record_comparison(&comparison, &labels);
```
joeykraut, line 81: i'd just break this out into a helper `record_comparison_for_source` or similar to be more readable

joeykraut (approved):

PR #88, opened 2025-01-18
auth-server: centralize label creation

sehyunc:
### Purpose
This PR bubbles up label creation for labels that are applied to all metrics of a request (e.g. request ID and key description). This ensures request ID can be used to link metrics + logs together since the same one is used. To this end, request ID is added as an attribute to bundle logs.

### Testing
- [x] Tested locally
- [ ] Tested in testnet

### DIFF
--- a/auth/auth-server/src/telemetry/helpers.rs
+++ b/auth/auth-server/src/telemetry/helpers.rs
@@ -22,12 +22,11 @@ use tracing::{info, warn};
 use crate::{
     error::AuthServerError,
     telemetry::labels::{
-        ASSET_METRIC_TAG, BASE_ASSET_METRIC_TAG, DECIMAL_CORRECTION_FIXED_METRIC_TAG,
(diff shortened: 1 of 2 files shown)


auth/auth-server/src/telemetry/helpers.rs
```diff
@@ -140,7 +136,7 @@ fn record_external_match_request_metrics(
     let quote_amount = order.get_quote_amount(fixed_point_price);
     let base_amount = order.get_base_amount(fixed_point_price);
 
-    record_volume_with_tags(&base_mint, base_amount, EXTERNAL_ORDER_BASE_VOLUME, labels);
+    record_volume_with_tags(&base_mint, base_amount, EXTERNAL_ORDER_BASE_VOLUME, &labels);
```
joeykraut, line 139: you've got clippy lints here and elsewhere

joeykraut (approved):
approved, but please take care of the linter errors

PR #92, opened 2025-01-18
auth-server: add gas cost comparison

sehyunc:
### Purpose
This PR adds methods to fetch gas price and ETH/USDC rate to estimate the gas cost of executing a quote.

### Testing
- [x] Tested locally
- [ ] Test in testnet


auth/auth-server/src/telemetry/quote_comparison/handler.rs
```diff
@@ -72,4 +63,66 @@ impl QuoteComparisonHandler {
...
+        // Convert wei to eth
+        let eth_string =
+            format_units(gas_price, "eth").map_err(|e| AuthServerError::Custom(e.to_string()))?;
+
+        eth_string.parse::<f64>().map_err(|e| AuthServerError::Custom(e.to_string()))
+    }
+
+    /// Fetches the current price of ETH in USDC
+    async fn fetch_eth_price(&self) -> Result<f64, AuthServerError> {
+        let eth = Token::from_ticker("WETH");
```
akirillo, line 119: can we leave a comment here calling out that the price reporter actually fetches the ETH (not WETH) price under the hood

auth/auth-server/src/telemetry/quote_comparison/mod.rs
```diff
@@ -10,9 +11,12 @@ const DECIMAL_TO_BPS: f64 = 10_000.0;
 
 /// Represents a single quote comparison between quotes from different sources
 pub struct QuoteComparison<'a> {
```
akirillo, line 13: First time I'm looking at this & it's out of scope for this PR / review, but why do we hold references (`&'a QuoteResponse`) instead of owned types (`QuoteResponse`) here? Generally find this to make things much less ergonomic due to managing lifetimes.

`QuoteResponse` is a pretty lightweight struct, it should be totally fine to clone it into `QuoteComparison` if ownership can't be taken directly

auth/auth-server/src/telemetry/quote_comparison/relayer_client.rs
akirillo, line 1: Looks like this file was copied over from the fee sweeper, which is sensible. However, since we're just relying on the price report route, we should go direct to the price reporter & just request the USDT-quoted price. This will take load off of the relayer, and since it's not the source of truth for prices it's unclear that we should continue to support the price report route on it. It should be simple enough to build an HTTP client for it (or even a single helper method for fetching a price)

auth/auth-server/src/telemetry/sources/mod.rs
```diff
@@ -7,6 +7,11 @@ use renegade_api::http::external_match::AtomicMatchApiBundle;
 use renegade_circuit_types::{order::OrderSide, Amount};
 use renegade_common::types::token::Token;
 
+/// The gas estimation to use if fetching a gas estimation fails
+const DEFAULT_GAS_ESTIMATION: u64 = 4_000_000; // 4m
```
akirillo, line 11: just out of curiosity, where did this default come from, and can we document that?

auth/auth-server/src/telemetry/sources/mod.rs
```diff
@@ -43,6 +40,49 @@ impl QuoteResponse {
...
+
+/// Converts the `AtomicMatchApiBundle` into a `QuoteResponse`.
+impl From<&AtomicMatchApiBundle> for QuoteResponse {
+    fn from(bundle: &AtomicMatchApiBundle) -> Self {
+        Self {
+            quote_mint: bundle.match_result.quote_mint.clone(),
+            base_mint: bundle.match_result.base_mint.clone(),
+            quote_amount: bundle.match_result.quote_amount,
+            base_amount: bundle.match_result.base_amount,
+            gas: DEFAULT_GAS_ESTIMATION,
```
akirillo, line 82: why don't we actually estimate the gas cost of the `settlement_tx` in the `AtomicMatchApiBundle`?

akirillo (approved):
have some questions & a couple nits, but nothing blocking

auth/auth-server/src/telemetry/sources/mod.rs
sehyunc: good point, we now use the estimate if it exists

auth/auth-server/src/telemetry/sources/mod.rs
sehyunc: comes from https://github.com/renegade-fi/renegade/blob/main/workers/api-server/src/http/external_match.rs/#L62, documented this

akirillo (approved):

PR #94, opened 2025-01-21
auth-server: record metrics for output net of gas and fees

sehyunc:
### Purpose
This PR records metrics net of both gas costs and fees.

### Testing
- [x] Tested locally
- [ ] Test in testnet

### DIFF
--- a/auth/auth-server/src/telemetry/quote_comparison/mod.rs
+++ b/auth/auth-server/src/telemetry/quote_comparison/mod.rs
@@ -63,4 +63,13 @@ impl<'a> QuoteComparison<'a> {
 
         calc_bps_diff(our_output_value_net_of_fee, source_output_value_net_of_fee)
     }
+
+    /// Calculate the net output value difference in basis points (bps).
+    /// Positive bps indicates our quote is better.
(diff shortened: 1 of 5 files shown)


auth/auth-server/src/telemetry/quote_comparison/mod.rs
```diff
@@ -62,4 +62,19 @@ impl<'a> QuoteComparison<'a> {
...
+        let our_output_value_net_of_gas =
+            self.our_quote.output_net_of_gas_and_fee(side, usdc_per_gas);
+        let source_output_value_net_of_gas =
+            self.source_quote.output_net_of_gas_and_fee(side, usdc_per_gas);
+
+        let diff_ratio = (our_output_value_net_of_gas - source_output_value_net_of_gas)
+            / source_output_value_net_of_gas;
+        let bps_diff = diff_ratio * DECIMAL_TO_BPS;
+
+        bps_diff as i32
```
joeykraut, line 78: same as the last PR, this should probably be an `f64` computation. Also, since it shows up so much, might be worth refactoring into a helper

joeykraut (approved):

PR #96, opened 2025-01-31
auth-server: handle_external_match: mutate calldata for gas sponsorship

akirillo:
This PR adds the ability to redirect the transaction in assembled quotes to a configured gas sponsorship context by specifying the `use_gas_sponsorship` (and optionally `refund_address`) query params.

### Testing
I tested this by invoking external matches on a locally-running auth server pointed against the testnet stack, trying buys/sells on both normal ERC20s and native ETH. I asserted that the gas sponsorship contract was being invoked as expected and refunding w/in 15k gas worth of ETH to the desired address. I tested w/ and w/o a refund address set, as well.

### TODO
- Add gas sponsorship to the direct match bundle route
- Gas sponsorship rate limiting
- Telemetry/monitoring around gas sponsorship

### DIFF
--- a/auth/auth-server/src/main.rs
+++ b/auth/auth-server/src/main.rs
@@ -20,7 +20,7 @@ pub(crate) mod schema;
 mod server;
 mod telemetry;
 
-use auth_server_api::API_KEYS_PATH;
+use auth_server_api::{ExternalQuoteAssemblyQueryParams, API_KEYS_PATH};
 use clap::Parser;
 use ethers::signers::LocalWallet;
 use renegade_arbitrum_client::{
@@ -99,11 +99,21 @@ pub struct Cli {
     #[clap(short, long, env = "RPC_URL")]
     rpc_url: String,
     /// The address of the darkpool contract
-    #[clap(short = 'a', long, env = "DARKPOOL_ADDRESS")]
+    #[clap(short, long, env = "DARKPOOL_ADDRESS")]
     darkpool_address: String,
     /// The URL of the price reporter
     #[arg(long, env = "PRICE_REPORTER_URL")]
     pub price_reporter_url: String,
+
+    // -------------------
+    // | Gas Sponsorship |
+    // -------------------
+    /// The address of the gas sponsor contract
+    #[clap(long, env = "GAS_SPONSOR_ADDRESS")]
+    gas_sponsor_address: String,
+    /// The auth private key used for gas sponsorship, encoded as a hex string
+    #[clap(long, env = "GAS_SPONSOR_AUTH_KEY")]
+    gas_sponsor_auth_key: String,
 
     // -------------
     // | Telemetry |
@@ -267,9 +277,10 @@ async fn main() {
         .and(warp::path::full())
         .and(warp::header::headers_cloned())
         .and(warp::body::bytes())
+        .and(warp::query::<ExternalQuoteAssemblyQueryParams>())
         .and(with_server(server.clone()))
-        .and_then(|path, headers, body, server: Arc<Server>| async move {
-            server.handle_external_quote_assembly_request(path, headers, body).await
+        .and_then(|path, headers, body, query_params, server: Arc<Server>| async move {
+            server.handle_external_quote_assembly_request(path, headers, body, query_params).await
         });
 
     let atomic_match_path = warp::path("v0")
--- a/auth/auth-server/src/server/handle_external_match.rs
+++ b/auth/auth-server/src/server/handle_external_match.rs
@@ -3,8 +3,15 @@
 //! At a high level the server must first authenticate the request, then forward
 //! it to the relayer with admin authentication
 
+use alloy_primitives::Address;
+use alloy_sol_types::{sol, SolCall};
+use auth_server_api::ExternalQuoteAssemblyQueryParams;
 use bytes::Bytes;
-use http::Method;
+use http::header::CONTENT_LENGTH;
+use http::{Method, Response};
+use renegade_arbitrum_client::abi::{
+    processAtomicMatchSettleCall, processAtomicMatchSettleWithReceiverCall,
+};
 use tracing::{info, instrument, warn};
 use warp::{reject::Rejection, reply::Reply};
 
@@ -15,6 +22,7 @@ use renegade_api::http::external_match::{
 use renegade_circuit_types::fixed_point::FixedPoint;
 use renegade_common::types::{token::Token, TimestampedPrice};
 
+use super::helpers::{gen_signed_sponsorship_nonce, get_selector};
 use super::Server;
 use crate::error::AuthServerError;
 use crate::telemetry::helpers::calculate_implied_price;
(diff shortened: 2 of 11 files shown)


auth/auth-server/src/main.rs
```diff
@@ -99,12 +99,22 @@ pub struct Cli {
...
     #[arg(long, env = "PRICE_REPORTER_URL")]
     pub price_reporter_url: String,
 
+    // -------------------
+    // | Gas Sponsorship |
+    // -------------------
+    /// The address of the gas sponsor contract
+    #[clap(long, env = "GAS_SPONSOR_ADDRESS")]
+    gas_sponsor_address: String,
+    /// The auth private key used for gas sponsorship
```
joeykraut, line 114: can we note here that we expect a hex encoding?

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -28,6 +39,37 @@ use crate::telemetry::{
...
+// | Types |
+// ---------
+
+// The ABI for gas sponsorship functions
+sol! {
+    function sponsorAtomicMatchSettle(bytes memory internal_party_match_payload, bytes memory valid_match_settle_atomic_statement, bytes memory match_proofs, bytes memory match_linking_proofs, address memory refund_address, uint256 memory nonce, bytes memory signature) external payable;
+    function sponsorAtomicMatchSettleWithReceiver(address receiver, bytes memory internal_party_match_payload, bytes memory valid_match_settle_atomic_statement, bytes memory match_proofs, bytes memory match_linking_proofs, address memory refund_address, uint256 memory nonce, bytes memory signature) external payable;
+}
+
+/// The query parameters accepted by the external quote assembly endpoint
```
joeykraut, line 60: i'd co-locate this with the other API types in the API crate

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -64,14 +106,40 @@ impl Server {
...
+        let key_desc = self.authorize_request(&auth_path, &headers, &body).await?;
         self.check_bundle_rate_limit(key_desc.clone()).await?;
 
         // Send the request to the relayer
-        let resp =
-            self.send_admin_request(Method::POST, path.as_str(), headers, body.clone()).await?;
+        let resp = if query_params.use_gas_sponsorship.unwrap_or(false) {
+            let original_resp =
+                self.send_admin_request(Method::POST, path.as_str(), headers, body.clone()).await?;
+
```
joeykraut, line 127: the else block is redundant here. we should move this line above the `if`

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -330,4 +398,140 @@ impl Server {
...
+        if external_match_resp.match_bundle.settlement_tx.gas().is_some() {
+            // If gas was estimated for the match, we need to re-estimate
+            // - there is extra overhead through the gas sponsor
+            // contract that must be accounted for.
+
+            let gas =
+                self.estimate_gas(external_match_resp.match_bundle.settlement_tx.clone()).await?;
+
+            external_match_resp.match_bundle.settlement_tx.set_gas(gas);
+        }
```
joeykraut, line 427: do we really need to do this? currently the relayer hardcodes this gas estimate, so i'd suggest we just don't bother for now and add this as a v1 if need be.

The concern here is that we're introducing another tx simulation in our critical path, and these endpoints need to be fast. I'd hate for the auth server to become the bottleneck.

We can just hardcode it here to the max gas we think they will pay given sponsorship; like 30k or something

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -330,4 +398,140 @@ impl Server {
...
+            .settlement_tx
+            .data()
+            .ok_or(AuthServerError::calldata_mutation("expected calldata"))?;
+
+        let selector: [u8; 4] = calldata
+            .as_ref()
+            .get(0..4)
+            .ok_or(AuthServerError::calldata_mutation("expected selector"))?
+            .try_into()
+            .unwrap();
```
joeykraut, line 462: i'd just break this out into a helper (`get_selector`), may read more fluidly

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -330,4 +398,140 @@ impl Server {
...
+
+        let gas_sponsor_calldata = match selector {
+            processAtomicMatchSettleCall::SELECTOR => {
```
joeykraut, line 465: i would also break out each of these match arms into helpers, less nested control flow = better to read

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -330,4 +398,140 @@ impl Server {
...
+        Ok(gas_sponsor_calldata)
+    }
+
+    /// Estimate the gas for the given external match transaction
+    async fn estimate_gas(&self, mut tx: TypedTransaction) -> Result<U256, AuthServerError> {
+        // To estimate gas without reverts, we would need to approve the ERC20 transfers
+        // due in the transaction before estimating. This is infeasible, so we mock the
+        // sender as the _darkpool itself_, which will automatically have an approval
+        // for itself. This can still fail if a transfer exceeds the darkpool's balance,
+        // in which case we fall back to the default gas estimation below
```
joeykraut, line 524: this doesn't work in mainnet FYI. hence my suggestion to just hardcode our estimation for now

auth/auth-server/src/server/helpers.rs
```diff
@@ -50,6 +53,45 @@ pub fn aes_decrypt(value: &str, key: &[u8]) -> Result<String, AuthServerError> {
     Ok(plaintext)
 }
 
+/// Generate a random nonce for gas sponsorship, signing it and the provided
+/// refund address
+pub fn generate_sponsorship_auth(
```
joeykraut, line 58: nit: maybe `sign_sponsorship_tx` or something like that? makes it clearer what is meant by "generating auth"

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -64,14 +106,40 @@ impl Server {
...
-            self.send_admin_request(Method::POST, path.as_str(), headers, body.clone()).await?;
+        let resp = if query_params.use_gas_sponsorship.unwrap_or(false) {
+            let original_resp =
+                self.send_admin_request(Method::POST, path.as_str(), headers, body.clone()).await?;
+
+            // If gas sponsorship is requested, mutate the calldata in the response
+            // to invoke the gas sponsor contract
+
+            info!("Redirecting match bundle through gas sponsor");
+            let refund_address = query_params
```
joeykraut, line 132: what if the sender does not know the refund address? we should allow this to default to `tx.origin` as discussed

joeykraut (commented):
overall looks good, have a few things to discuss

auth/auth-server/src/server/handle_external_match.rs
akirillo: i don't mind sticking with the default estimation, but i'm not sure it's a good idea to use the expected post-sponsorship gas amount. this field is used to set the tx gas limit which, if it's too low, will cause reverts

auth/auth-server/src/server/handle_external_match.rs
akirillo: this default is handled in the contract. the refund address can be omitted from the query params, in which case it gets defaulted to the zero address here. in the contract, a zero address results in us using `tx.origin`

auth/auth-server/src/server/handle_external_match.rs
joeykraut: if users are to make use of this feature they will be overwriting the gas limit anyways. we may as well give an accurate estimate. overestimating here by two orders of magnitude is far worse

akirillo: if we want to give an accurate estimate, I'd rather have that be a separate field on the API response than to co-opt this field on the tx, and leave the tx gas field untouched. how does that sound?

joeykraut: why? it makes no sense to quote 4m gas (the relayer default) in the `tx.gas` field when it will cost <1% of that.

akirillo: because the field is intended to be used to set the gas limit for a transaction. it feels like an indirect & obfuscated way to communicate the expected out-of-pocket gas costs to the user where we could be much clearer.

in reality, the tx will use the same amount of gas, we will just refund the appropriate ether amount to the user.

joeykraut: gotcha, okay that's a fair way to think about it, sgtm. 

I don't think we need an extra field, unless someone asks for it

joeykraut (approved):

PR #98, opened 2025-02-01
auth-server: server: rate_limiter: initial gas sponsorship rate limiter impl

akirillo:
This PR lays out the implementation of the `GasSponsorshipRateLimiter`, modeled after the `UserRateLimiter`. We can't, however, use then single-token-based rate limiting implemented by the `ratelimit` as we do in the `UserRateLimiter`. As such, we make a custom implementation which has a maximum $ value of sponsorship for each bucket, from which each successful sponsorship subtracts.

We allow sponsorship to occur as long as a user's bucket is not empty, i.e. we allow the gas sponsorship budget to be exceeded for requests made before the bucket value is decremented past 0. This is a deliberate decision to err on the side of liveness, and it makes the implementation simpler. The alternative would be a preemptive subtraction of expected sponsorship costs at the time the rate limit is checked, with a refund in the case of gas sponsorship not occurring (e.g. bundle didn't settle). This would be similar to what we do for the bundle rate limiter. However, we see bundles _not_ settle quite often at the moment, so this behavior may result in too stifling of a rate limit. Additionally, this would include the complexity & inaccuracy of computing expected gas sponsorship amounts at the time of request, and price drift in the value of ETH between when the rate limit is checked to when the rate limited amount is refunded to the bucket in the case of the bundle not settling.

This will be integrated into the auth server code paths in the following PR.


auth/auth-server/src/server/rate_limiter/gas_sponsorship_rate_limiter.rs
```diff
@@ -0,0 +1,138 @@
...
+        // Potentially refill the bucket, if it is due
+        self.try_refill();
+
+        // Return true if there is any remaining value
+        self.remaining_value > 0.0
+    }
+
+    /// Record a gas sponsorship value.
+    ///
+    /// If the bucket does not have any remaining value, this method will do
```
joeykraut, line 112: nit: looks like this sentence was cut off

auth/auth-server/src/server/rate_limiter/user_rate_limiter.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// If no token is available (rate limit reached), this method returns
+    /// false, otherwise true
+    pub async fn check(&self, user_id: String) -> bool {
+        let mut map = self.buckets.lock().await;
+        let entry = map.entry(user_id).or_insert_with(|| self.new_rate_limiter());
+        entry.try_wait().is_ok()
+    }
+
+    /// Increment the number of tokens available to a given user
+    #[allow(unused_must_use)]
```
joeykraut, line 64: do we need this clippy lint allow still?

joeykraut (approved):

auth/auth-server/src/server/rate_limiter/user_rate_limiter.rs
akirillo: yes we do, this is because we deliberately ignore the potential `ratelimit` error in exceeding the maximum bucket tokens

PR #99, opened 2025-02-01
auth-server: server: handel_external_match: use gas sponsorship rate limiter

akirillo:
This PR integrates usage of the gas sponsorship rate limiter into the external match paths in the auth server. We do this by checking the rate limit when a request for a sponsored match comes in, and consuming the user's gas sponsorship budget if they settle the bundle by fetching the ETH price and the amount of gas sponsored for the request.

### Testing
- [x] Tested w/ local auth server against testnet using a low rate limit, confirmed that rate limit was exceeded & refilled accordingly

### DIFF
--- a/auth/auth-server/src/server/handle_external_match.rs
+++ b/auth/auth-server/src/server/handle_external_match.rs
@@ -5,10 +5,13 @@
 
 use alloy_primitives::Address;
 use alloy_sol_types::{sol, SolCall};
-use auth_server_api::GasSponsorshipQueryParams;
+use auth_server_api::{ExternalMatchResponse, GasSponsorshipQueryParams};
 use bytes::Bytes;
+use ethers::contract::abigen;
+use ethers::types::transaction::eip2718::TypedTransaction;
+use ethers::types::U256;
 use http::header::CONTENT_LENGTH;
-use http::{Method, Response};
+use http::{Method, Response, StatusCode};
 use renegade_arbitrum_client::abi::{
     processAtomicMatchSettleCall, processAtomicMatchSettleWithReceiverCall,
 };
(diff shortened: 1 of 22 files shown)


auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -426,44 +439,60 @@ impl Server {
...
+        let mut relayer_external_match_resp: RelayerExternalMatchResponse =
             serde_json::from_slice(resp.body()).map_err(AuthServerError::serde)?;
 
-        let gas_sponsor_calldata =
-            self.generate_gas_sponsor_calldata(&external_match_resp, refund_address)?.into();
+        relayer_external_match_resp.match_bundle.settlement_tx.set_to(self.gas_sponsor_address);
+
+        let refund_address = query_params.get_refund_address()?;
+        let is_sponsored = query_params.use_gas_sponsorship.unwrap_or(false)
+            && self.check_gas_sponsorship_rate_limit(key_desc.clone()).await;
```
joeykraut, line 457: I always like to check rate limits at the top of a handler so that no other work needs to be done if the rate limit is enforced

auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -426,44 +439,60 @@ impl Server {
...
         let mut new_resp = warp::http::Response::new(body);
         *new_resp.status_mut() = resp.status();
 
-        // Remove the old content length header so we don't overwrite the new one,
-        // it is incorrect since the calldata was mutated
         let mut headers = resp.headers().clone();
-        headers.remove(CONTENT_LENGTH);
 
+        // Remove the old content length header since it is no longer correct
+        headers.remove(CONTENT_LENGTH);
```
joeykraut, line 495: you should add this back to be `body.len()` o/w some http clients will be unable to fetch the response

joeykraut (approved):

PR #101, opened 2025-02-02
auth-server: server: handle_external_match: add gas sponsorship metrics

akirillo:
This PR adds the recording of a metric tracking the $ value spent in gas sponsorship for settled matches.

The second commit in this PR is purely cosmetic and moves around various gas sponsorship helpers into a single module for cleanliness. Only the first commit needs to be reviewed for soundness.

### DIFF
--- a/auth/auth-server/src/server/handle_external_match.rs
+++ b/auth/auth-server/src/server/handle_external_match.rs
@@ -8,8 +8,8 @@ use alloy_sol_types::{sol, SolCall};
 use auth_server_api::{ExternalMatchResponse, GasSponsorshipQueryParams};
 use bytes::Bytes;
 use ethers::contract::abigen;
-use ethers::types::transaction::eip2718::TypedTransaction;
-use ethers::types::U256;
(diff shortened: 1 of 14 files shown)


auth/auth-server/src/server/handle_external_match.rs
```diff
@@ -640,12 +637,45 @@ impl Server {
...
+            let eth_price: f64 = self
+                .price_reporter_client
+                .get_eth_price()
+                .await
+                .map_err(AuthServerError::custom)?;
+
+            let gas_sponsorship_value = eth_price * gas_cost_eth;
+
+            self.record_gas_sponsorship_rate_limit(key, gas_sponsorship_value).await?;
+
```
sehyunc: Seems like we can just invoke self.rate_limiter.record_gas_sponsorship() here, but maybe you have future need to abstract some logic away?

sehyunc (approved):
LGTM

auth/auth-server/src/server/handle_external_match.rs
akirillo: fair point, needless indirection

PR #103, opened 2025-02-03
funds-manager: custody_client: gas_sponsor: add gas sponsor refill route

akirillo:
This PR adds the `/custody/gas/refill-gas-sponsor` route to the funds manager, used to refill the gas sponsor's ETH balance. This is necessary, i.e. the `/custody/gas/withdraw-gas` route can't be used, since the `receiveEth` method on the gas sponsor must be invoked.

### Testing
This was tested by running a local funds manager against the testnet stack & invoking the endpoint.


joeykraut (approved):
clean

PR #105, opened 2025-02-04
auth-server: telemetry: gracefully handle quote comparison errors

akirillo:
This PR ensures that we  gracefully handle errors in quote comparison recording so that we don't panic the thread.

### DIFF
--- a/auth/auth-server/src/server/handle_external_match/mod.rs
+++ b/auth/auth-server/src/server/handle_external_match/mod.rs
@@ -247,7 +247,8 @@ impl Server {
         }
 
         // Record metrics
-        record_external_match_metrics(&order, match_resp.match_bundle, &labels, did_settle).await?;
+        record_external_match_metrics(&order, &match_resp.match_bundle, &labels, did_settle)
+            .await?;
 
         Ok(())
     }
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -247,7 +240,15 @@ impl Server {
...
         // Record metrics
-        record_external_match_metrics(&order, match_resp.match_bundle, &labels, did_settle).await?;
+        record_external_match_metrics(&order, &match_resp.match_bundle, &labels, did_settle)
+            .await?;
+
+        // Record quote comparisons before settlement, if enabled
+        if let Some(quote_metrics) = &self.quote_metrics {
+            quote_metrics
+                .record_quote_comparison(&match_resp.match_bundle, labels.as_slice())
+                .await;
```
sehyunc, line 250: We intentionally record quote comparison metrics before awaiting settlement to compare against alternative quotes as early as possible for accuracy, as solvers are likely comparing quotes as soon as they receive them. Can we keep the ordering but not panic on errors within `record_quote_comparison`?

sehyunc (changes requested):
Seems like your recent changes are heading in this direction

auth/auth-server/src/server/handle_external_match/mod.rs
akirillo: yup ack, my subsequent fixes allow for this, will do

sehyunc (approved):
lgtm, thanks

PR #110, opened 2025-03-06
auth-server: handle_external_match: gas_sponsorship: compute conversion rate

akirillo:
This PR implements the necessary changes to invoke external matches sponsored in-kind through the gas sponsor. Primarily, this is composed of an additional query parameter `refund_native_eth` - which defaults to `false`, implying that in-kind gas sponsorship is the default - and the logic for computing a conversion rate for in-kind sponsorship.

### TODO
- [x] Rate limiting of in-kind refunds

### Testing
- [x] Conversion rate calculation, unit tests
- [x] Testing against testnet
    - [x] Simple buy w/ in-kind refund
    - [x] Simple sell w/ in-kind refund
    - [x] Native ETH buy w/ in-kind refund
    - [x] Native ETH sell w/ in-kind refund
    - [x] In-kind refund w/ refund address set
    - [x] In-kind refund w/ receiver set
    - [ ] Native ETH refund w/ receiver set (blocked on SDK updates)

### DIFF
--- a/auth/auth-server-api/src/lib.rs
+++ b/auth/auth-server-api/src/lib.rs
@@ -52,12 +52,16 @@ pub struct SponsoredMatchResponse {
 /// The query parameters used for gas sponsorship
 #[derive(Debug, Serialize, Deserialize)]
 pub struct GasSponsorshipQueryParams {
-    /// Whether to use gas sponsorship for the external match
-    pub use_gas_sponsorship: Option<bool>,
-    /// The address to refund gas to
+    /// Whether to omit gas sponsorship for the external match.
+    /// Defaults to `false`, meaning gas sponsorship is used.
+    pub omit_sponsorship: Option<bool>,
+    /// The address to refund gas to.
+    /// In the case of a native ETH refund, defaults to `tx::origin`.
+    /// In the case of an in-kind refund, defaults to the receiver.
     pub refund_address: Option<String>,
     /// Whether to provide the gas refund in terms of native ETH,
-    /// as opposed to the buy-side token
+    /// as opposed to the buy-side token.
+    /// Defaults to `false`, meaning the buy-side token is used.
     pub refund_native_eth: Option<bool>,
 }
 
--- a/auth/auth-server-api/src/lib.rs
+++ b/auth/auth-server-api/src/lib.rs
@@ -56,6 +56,9 @@ pub struct GasSponsorshipQueryParams {
     pub use_gas_sponsorship: Option<bool>,
     /// The address to refund gas to
     pub refund_address: Option<String>,
+    /// Whether to provide the gas refund in terms of native ETH,
+    /// as opposed to the buy-side token
+    pub refund_native_eth: Option<bool>,
 }
 
 impl GasSponsorshipQueryParams {
--- a/auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
+++ b/auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
@@ -6,6 +6,7 @@
 use alloy_primitives::{Address, Bytes as AlloyBytes, U256 as AlloyU256};
 use alloy_sol_types::{sol, SolCall};
 use auth_server_api::SponsoredMatchResponse;
+use bigdecimal::One;
 use bigdecimal::{num_bigint::BigInt, BigDecimal, FromPrimitive};
 use bytes::Bytes;
 use ethers::contract::abigen;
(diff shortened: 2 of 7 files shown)


auth/auth-server-api/src/lib.rs
```diff
@@ -56,6 +56,9 @@ pub struct GasSponsorshipQueryParams {
     pub use_gas_sponsorship: Option<bool>,
     /// The address to refund gas to
     pub refund_address: Option<String>,
+    /// Whether to provide the gas refund in terms of native ETH,
+    /// as opposed to the buy-side token
+    pub refund_native_eth: Option<bool>,
```
joeykraut, line 61: can we note in this struct which values these fields take in default? For posterity, and to make sure we're on the same page, those should be true and false for `use_gas_sponsorship` and `refund_native_eth` respectively, and `refund_address` is `tx::origin` if the refund is a native eth refund, otherwise its the match receiver, right?

auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
```diff
@@ -42,18 +46,90 @@ abigen!(
     ]"#
 );
 
+impl sponsorAtomicMatchSettleWithRefundOptionsCall {
+    /// Create a `sponsorAtomicMatchSettleWithRefundOptions` call from
+    /// `processAtomicMatchSettle` calldata
+    pub fn from_process_atomic_match_settle_calldata(
```
joeykraut, line 52: clean

auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
```diff
@@ -84,12 +169,94 @@ impl Server {
         Ok(())
     }
 
+    /// Fetch the conversion rate from ETH to the buy-side token in the trade
+    /// from the price reporter, if necessary.
+    /// The conversion rate is in units of
+    /// `token/wei * 10^CONVERSION_RATE_SCALE`
+    #[allow(clippy::unused_async)]
+    async fn maybe_fetch_conversion_rate(
```
joeykraut, line 177: can you time the execution of a gas sponsored client request using e.g. the rust client? I want to know how much this slows down the critical path. We may need to be more clever if it's significant, for example by streaming ETH prices from the price reporter to a local client. And as mentioned below we don't need to fetch the base token

auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
```diff
@@ -84,12 +169,94 @@ impl Server {
...
+        // is native ETH, we don't need to get a conversion rate
+        if refund_native_eth || native_eth_buy {
+            return Ok(None);
+        }
+
+        // Get ETH price
+        let eth_price_f64 = self.price_reporter_client.get_eth_price().await?;
+
+        // Get TOKEN price
+        let buy_token_price_f64 = self.price_reporter_client.get_binance_price(buy_mint).await?;
```
joeykraut, line 195: why don't we just use the price we're already quoting at in the match? no need to make a second request

auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
```diff
@@ -84,12 +169,94 @@ impl Server {
...
+
+        // Decimal-adjust the rate to represent (smallest-denomination) *units* of TOKEN
+        // per ETH
+        let adjustment: BigDecimal = BigInt::from(10).pow(buy_token_decimals).into();
+        let conversion_rate_adjusted = conversion_rate * adjustment;
+
+        // Convert the scaled rate to a U256. We can use the `BigInt` component of the
+        // `BigDecimal` directly because we round to 0 digits after the decimal.
+        let (conversion_rate_bigint, _) =
+            conversion_rate_adjusted.round(0 /* round_digits */).into_bigint_and_scale();
```
joeykraut, line 239: i'm guessing we don't need an eth -> wei decimal adjustment because we've set the fixed point precision here to be the same as the decimals on WETH?

If that's the case can you note that in the comments, took me a minute to realize what was going on

joeykraut (approved):

auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
akirillo: fair point, i'll time it in testnet and if it's significant i'll put that change in a follow-up PR

auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
akirillo: This is correct. It's equivalent to saying our desired conversion rate is simply in terms of nominal units of the buy-side token per 1 whole ETH, rounded to the nearest integer. I found that simpler to reason about than considering the conversion rate as a separate fixed-point number with its own precision, so I've updated all the comments to refer to the units of the conversion rate in this way

PR #111, opened 2025-03-07
auth-server: handle_external_match, telemetry: record in-kind gas sponsorship value

akirillo:
This PR updates the way in which we record the value of gas sponsorship, both for rate limiting and telemetry, to account for in-kind sponsorship.

### Testing
- [x] Test rate limiting against testnet
- [x] Test metrics against testnet

### DIFF
--- a/auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
+++ b/auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
@@ -3,30 +3,31 @@
 //! At a high level the server must first authenticate the request, then forward
 //! it to the relayer with admin authentication
 
-use alloy_primitives::{Address, Bytes as AlloyBytes, U256 as AlloyU256};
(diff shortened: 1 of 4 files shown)


auth/auth-server/src/server/handle_external_match/gas_sponsorship.rs
```diff
@@ -352,20 +348,37 @@ impl Server {
...
+                // The zero address indicates that the gas was sponsored via native ETH.
+                // In this case, we can substitute the WETH token, as it has the same
+                // number of decimals and yields the same price from the price reporter.
+                Token::from_ticker("WETH")
+            } else {
+                Token::from_addr(&format!("{:#x}", token_addr))
+            };
+
+            // Get the price of the token in USD
+            let price: f64 = self
```
joeykraut, line 364: same as my comment in the last pr, we can probably just use our quoted price both to compute the conversion rate and to record metrics

joeykraut (approved):

PR #112, opened 2025-03-08
auth-server: price_reporter_client: new price reporter client module w/ price stream

akirillo:
This PR augments the `PriceReporterClient` in the auth server with a ETH price stream via a websocket subscription to the price reporter.

The implementation mirrors that of the price reporter websockets in the relayer, but simplified as it only manages one price.

This required some moving around of code as the price reporter client used to be submodule of the telemetry module.

### Testing
- [x] Testing external matches against testnet

Testing shows that the time to fetch an ETH price in the critical path of assembling a quote drops from ~400ms to ~100 microseconds


auth/auth-server-api/src/lib.rs
```diff
@@ -52,9 +52,9 @@ pub struct SponsoredMatchResponse {
 /// The query parameters used for gas sponsorship
 #[derive(Debug, Serialize, Deserialize)]
 pub struct GasSponsorshipQueryParams {
-    /// Whether to omit gas sponsorship for the external match.
```
joeykraut, line 55: might be simpler to just have an `enum GasSponsorshipType` with options for `None, InKind, NativeEth` here rather than two bools. you'd have to make this backwards compatible, so up to you if that's worth it now, but for future reference that would make a cleaner api

joeykraut (approved):

auth/auth-server-api/src/lib.rs
akirillo: Definitely more ergonomic code, but defining the struct in this way makes it directly deserializable via `serde_urlencoded` which makes using this in the `warp` stack simple

joeykraut: i'd say the bigger thing is that it's a cleaner API, some of our integrations call the API directly and having to reason about combinations of booleans is not ideal

PR #115, opened 2025-03-09
auth-server: gas_sponsorship: gas_estimation: core gas estimation logic

akirillo:
This PR scaffolds the `gas_estimation` module and implements the core gas estimation logic, which for the moment is very simple. We use a constant estimate of both the amount of L2 gas needed to execute the external match, and the calldata size in bytes. We use these values along with a real-time sampling of the L2 basefee and estimated amount of L1 gas used from the [`NodeInterface`](https://docs.arbitrum.io/build-decentralized-apps/nodeinterface/overview) precompile to compute an estimated gas cost according to the instructions [here](https://docs.arbitrum.io/build-decentralized-apps/how-to-estimate-gas).

This functionality is not yet used so clippy lints are expected.

### DIFF
--- a/auth/auth-server/src/server/handle_external_match/gas_sponsorship/gas_estimation.rs
+++ b/auth/auth-server/src/server/handle_external_match/gas_sponsorship/gas_estimation.rs
@@ -1,0 +1,74 @@
+//! Gas estimation for gas sponsorship
+
+use alloy_primitives::hex;
+use ethers::{
+    contract::abigen,
+    types::{Address, H160, U256},
+};
+use rand::{thread_rng, RngCore};
+
+use crate::{error::AuthServerError, server::Server};
+
+// -------------
+// | Constants |
+// -------------
+
+/// The estimated gas cost of L2 execution for an external match.
+/// From https://github.com/renegade-fi/renegade/blob/main/workers/api-server/src/http/external_match.rs/#L62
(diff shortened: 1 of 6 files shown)


auth/auth-server/src/server/handle_external_match/gas_sponsorship/gas_estimation.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use crate::{error::AuthServerError, server::Server};
+
+// -------------
+// | Constants |
+// -------------
+
+/// The estimated gas cost of L2 execution for an external match.
+/// From https://github.com/renegade-fi/renegade/blob/main/workers/api-server/src/http/external_match.rs/#L62
+pub const ESTIMATED_L2_GAS: u64 = 4_000_000; // 4m
```
joeykraut, line 18: for the record this was an overestimate when I originally made it, rounding up to the nearest 1m. We can leave it like is to keep a buffer, but that may obviate the need to overestimate the rebate elsewhere, FYI

joeykraut (approved):
very nice

auth/auth-server/src/server/handle_external_match/gas_sponsorship/gas_estimation.rs
akirillo: Yup that makes sense. I'm happy having it be an overestimate especially since the true value will drift from the time of sampling, but I will add a comment mentioning this explicitly

PR #116, opened 2025-03-10
auth-server: gas_estimation: create gas cost sampler module

akirillo:
This PR introduces the `gas_cost_sampler` module, responsible for updating a shared estimate of the estimated gas cost to perform an external match on a fixed interval. This estimate will be used for gas sponsorship, since it is a mostly static value that doesn't need to be computed in the critical path of serving a bundle.

This code is not yet integrated into the external match handler so clippy lints are expected.


auth/auth-server/src/server/gas_estimation/gas_cost_sampler.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /// Get the latest estimate of the gas cost for an external match
+    pub async fn get_latest_estimate(&self) -> U256 {
+        self.latest_estimate.read().await.clone()
+    }
+
+    /// Estimate the gas cost, in wei, of an external match.
+    /// This calculation was taken from https://docs.arbitrum.io/build-decentralized-apps/how-to-estimate-gas
+    async fn estimate_external_match_gas_cost(&self) -> Result<(), String> {
```
akirillo, line 84: This function was directly copied over from gas_sponsorship/gas_estimation.rs

auth/auth-server/src/server/gas_estimation/gas_cost_sampler.rs
```diff
@@ -0,0 +1,116 @@
...
+        function gasEstimateL1Component(address to, bool contractCreation, bytes calldata data) external payable returns (uint64 gasEstimateForL1, uint256 baseFee, uint256 l1BaseFeeEstimate)
+    ]"#
+);
+
+/// A lightweight worker that periodically samples an estimate of the gas cost
+/// for an external match
+#[derive(Clone)]
+pub struct GasCostSampler {
+    /// The latest estimate of the gas cost for an external match
+    latest_estimate: Arc<RwLock<U256>>,
```
joeykraut, line 40: `AtomicCell` has a more convenient interface for this sort of access pattern, and will use a `RwLock` under the hood for types like `U256`. This is fine, but good to know for the future

joeykraut (approved):

PR #117, opened 2025-03-10
auth-server: gas_sponsorship: use exact refund amount

akirillo:
This PR lays the groundwork for sponsoring gas using exact token amounts. Namely, the changes here are:
1. Preempt contract interface changes: we now expect the contracts to directly receive a `refund_amount` for gas sponsorship
2. Use the new `gas_estimation` module combined with the existing conversion logic to compute the refund amount in the appropriate token
3. Configure abstraction boundaries to lift computation of the refund amount out of the calldata construction code path. This will be necessary for computing the amount during the "quote" stage of the "quote-assemble" flow, the changes for which will come in a later PR

### Testing
- [x] Assert that gas estimation module is correctly estimating gas prices on the expected interval while the server runs

### DIFF
--- a/auth/auth-server/src/server/handle_external_match/gas_sponsorship/mod.rs
+++ b/auth/auth-server/src/server/handle_external_match/gas_sponsorship/mod.rs
@@ -24,7 +24,8 @@ use renegade_api::http::external_match::{AtomicMatchApiBundle, ExternalMatchResp
 
 use super::Server;
 use crate::server::helpers::{
-    gen_signed_sponsorship_nonce, get_nominal_buy_token_price, get_selector,
+    ethers_u256_to_alloy_u256, gen_signed_sponsorship_nonce, get_nominal_buy_token_price,
+    get_selector,
 };
 use crate::telemetry::helpers::record_gas_sponsorship_metrics;
 use crate::{error::AuthServerError, server::helpers::ethers_u256_to_bigdecimal};
@@ -35,7 +36,7 @@ use crate::{error::AuthServerError, server::helpers::ethers_u256_to_bigdecimal};
 
 // The ABI for gas sponsorship functions
 sol! {
-    function sponsorAtomicMatchSettleWithRefundOptions(address receiver, bytes internal_party_match_payload, bytes valid_match_settle_atomic_statement, bytes match_proofs, bytes match_linking_proofs, address refund_address, uint256 nonce, bool refund_native_eth, uint256 conversion_rate, bytes signature) external payable;
+    function sponsorAtomicMatchSettleWithRefundOptions(address receiver, bytes internal_party_match_payload, bytes valid_match_settle_atomic_statement, bytes match_proofs, bytes match_linking_proofs, address refund_address, uint256 nonce, bool refund_native_eth, uint256 sponsorship_amount, bytes signature) external payable;
 }
 
 // The ABI for gas sponsorship events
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/server/handle_external_match/gas_sponsorship/mod.rs
```diff
@@ -123,50 +126,59 @@ impl sponsorAtomicMatchSettleWithRefundOptionsCall {
...
-        resp.headers_mut().insert(CONTENT_LENGTH, body.len().into());
-        *resp.body_mut() = body;
+        let conversion_rate =
+            self.maybe_fetch_conversion_rate(match_result, refund_native_eth).await?;
 
-        Ok(())
+        let estimated_gas_cost = ethers_u256_to_alloy_u256(self.get_gas_cost_estimate().await);
+
+        let refund_amount = if let Some(conversion_rate) = conversion_rate {
+            let wei_in_eth = ethers_u256_to_alloy_u256(WEI_IN_ETHER);
```
joeykraut, line 175: nit: why not just constantize this ourselves as 1e18, and avoid paying the conversion price every time

auth/auth-server/src/server/handle_external_match/gas_sponsorship/mod.rs
```diff
@@ -123,50 +126,59 @@ impl sponsorAtomicMatchSettleWithRefundOptionsCall {
...
         refund_address: AlloyAddress,
         refund_native_eth: bool,
-    ) -> Result<(), AuthServerError> {
-        let mut relayer_external_match_resp: ExternalMatchResponse =
-            serde_json::from_slice(resp.body()).map_err(AuthServerError::serde)?;
-
-        relayer_external_match_resp.match_bundle.settlement_tx.set_to(self.gas_sponsor_address);
+        refund_amount: Option<AlloyU256>,
+    ) -> Result<SponsoredMatchResponse, AuthServerError> {
+        external_match_resp.match_bundle.settlement_tx.set_to(self.gas_sponsor_address);
```
joeykraut, line 137: for my own understanding, do we need to set this back to the darkpool if the rate limit has been exceeded? or should calling through the gas sponsor still work?

auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -174,34 +195,52 @@ impl Server {
...
-        // requested sponsorship, regardless of whether they are rate-limited.
-        if sponsorship_requested {
+        let external_match_resp: ExternalMatchResponse =
+            serde_json::from_slice(resp.body()).map_err(AuthServerError::serde)?;
+
+        // Whether or not sponsorship was requested, we return a
+        // `SponsoredMatchResponse`. This simply has one extra field,
+        // `is_sponsored`, which we expect clients to ignore if they did not
+        // request sponsorship.
+        let sponsored_match_resp = if sponsorship_requested {
```
joeykraut, line 205: can this be in a helper given that we do the same control flow above?

joeykraut (approved):

auth/auth-server/src/server/handle_external_match/gas_sponsorship/mod.rs
akirillo: calling through the gas sponsor will still work! the calldata ends up being unchanged, and the gas sponsor exposes the same unsponsored methods as the darkpool, so the unchanged calldata will execute just fine

PR #120, opened 2025-03-12
auth-server: handle_external_match, gas_sponsorship: update external quote w/ sponsorship info

akirillo:
This PR scaffolds the process by which gas sponsorship will be applied to requested quotes. We update the returned quote if necessary, adding the gas sponsorship info to the response object. This is useful not only to clients, but also so that we can reconstruct the original relayer-signed quote during assembly. For this reason we sign the sponsorship info in the response.

In this PR, we also move around a bit of code to make for better-packaged helpers. This is in anticipation of the assembly endpoint using the gas sponsorship info from the quote. It's worth noting that, if we want to maintain perfect backwards compatibility, we would still need to support requesting gas sponsorship at the assembly step. However this will be handled as an edge case until it is deprecated, and the structure of the helpers should not reflect that. 

Of note, we now do _not_ invoke the gas sponsor contract if the gas sponsorship rate limit was exceeded. We previously would, though this makes no difference in the actual match settlement.

### TODO
Sponsored quote assembly will come in a follow up PR.

### DIFF
--- a/auth/auth-server-api/src/lib.rs
+++ b/auth/auth-server-api/src/lib.rs
@@ -7,7 +7,9 @@
 #![feature(trivial_bounds)]
 
 use alloy_primitives::Address;
-use renegade_api::http::external_match::AtomicMatchApiBundle;
+use renegade_api::http::external_match::{
+    AssembleExternalMatchRequest, AtomicMatchApiBundle, ExternalQuoteResponse,
+};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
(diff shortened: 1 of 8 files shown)


auth/auth-server/src/server/handle_external_match/gas_sponsorship/contract_interaction.rs
akirillo, line 1: Everything in this file was copied over directly from `gas_sponsorship/mod.rs`

auth/auth-server/src/server/handle_external_match/gas_sponsorship/refund_calculation.rs
```diff
@@ -0,0 +1,125 @@
...
+        };
+
+        Ok(refund_amount)
+    }
+
+    /// Fetch the conversion rate from ETH to the buy-side token in the trade
+    /// from the price reporter, if necessary.
+    /// The conversion rate is in terms of nominal units of TOKEN per whole ETH.
+    #[allow(clippy::unused_async)]
+    async fn maybe_fetch_conversion_rate(
```
akirillo, line 56: this method was directly copied over from `gas_sponsorship/mod.rs`, the other 2 in this file are new/updated and should be reviewed

auth/auth-server/src/server/handle_external_match/gas_sponsorship/contract_interaction.rs
akirillo, line 1: Everything in this file was copied over directly from `gas_sponsorship/mod.rs`

auth/auth-server-api/src/lib.rs
```diff
@@ -40,6 +42,50 @@ pub struct CreateApiKeyRequest {
...
+/// An external quote response from the auth server, potentially with
+/// gas sponsorship info
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct SponsoredQuoteResponse {
+    /// The external quote response from the relayer, potentially updated to
+    /// reflect the post-sponsorship price and receive amount
+    #[serde(flatten)]
+    pub external_quote_response: ExternalQuoteResponse,
+    /// The signed sponsorship info, if sponsorship was requested
+    pub signed_sponsorship_info: Option<SignedSponsorshipInfo>,
```
joeykraut, line 54: can we call this `signed_gas_sponsorship_info` throughout this file? That will make it clear to an API user exactly _what_ is being sponsored

auth/auth-server-api/src/lib.rs
```diff
@@ -40,6 +42,50 @@ pub struct CreateApiKeyRequest {
...
+
+/// A request to assemble a potentially sponsored quote into a settlement bundle
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct AssembleSponsoredMatchRequest {
+    /// The request to assemble the external match
+    #[serde(flatten)]
+    pub assemble_external_match_request: AssembleExternalMatchRequest,
+    /// The gas sponsorship info associated with the quote,
+    /// if sponsorship was requested
+    pub gas_sponsorship_info: Option<SignedSponsorshipInfo>,
```
joeykraut, line 86: we should probably name this the same field as in the quote response. I like this name better (it's not important that `signed` be in the name) so i'd suggest copying this name to the `SponsoredQuoteResponse`

auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -197,45 +169,95 @@ impl Server {
...
+
+        // Parse query params
+        let (sponsorship_requested, refund_address, refund_native_eth) =
+            query_params.get_or_default();
+
+        // Check gas sponsorship rate limit
+        let gas_sponsorship_rate_limited =
+            !self.check_gas_sponsorship_rate_limit(key_description).await;
+
+        let sponsor_match = !gas_sponsorship_rate_limited && sponsorship_requested;
```
joeykraut, line 193: i think it makes logical sense to check these conditions before deserializing the quote response

joeykraut (approved):

auth/auth-server/src/server/handle_external_match/mod.rs
akirillo: sure if it helps readability, but the quote response needs to be deserialized anyway

PR #121, opened 2025-03-12
auth-server: handle_external_match: sponsored quote assembly

akirillo:
This PR implements the process for sponsored quote assembly, in which we authenticate the re-submitted sponsorship info, reconstruct the original signed quote from the relayer to proxy the bundle request, and apply gas sponsorship to the resulting bundle accordingly.

We maintain backwards compatibility for the assembly endpoint, which currently supports requesting gas sponsorship. In the new quote-assemble flow, this doesn't make sense, as we intend for sponsorship to be requested at the quote stage and applied during assembly. So we enable gas sponsorship to be requested during assembly iff it wasn't requested during quoting.

### Testing
- [ ] Test against testnet


auth/auth-server/src/server/handle_external_match/gas_sponsorship/refund_calculation.rs
```diff
@@ -122,4 +122,31 @@ impl Server {
...
+                (quote.match_result.base_amount - refund_amount, quote.match_result.quote_amount)
+            },
+            OrderSide::Sell => {
+                (quote.match_result.base_amount, quote.match_result.quote_amount - refund_amount)
+            },
+        };
+
+        let base_amt_f64 = base_amount as f64;
+        let quote_amt_f64 = quote_amount as f64;
+        let price = quote_amt_f64 / base_amt_f64;
```
joeykraut, line 143: have you checked that this works? wondering if this accurately recovers the price given `fp64`'s precision. we need the _exact_ price for the relayer to validate the quote signature

auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -209,23 +246,99 @@ impl Server {
...
+            if sponsorship_info.requires_quote_update() {
+                // Reconstruct original signed quote
+                let quote = &mut assemble_sponsored_match_req
+                    .assemble_external_match_request
+                    .signed_quote
+                    .quote;
+
+                self.remove_sponsorship_from_quote(quote, sponsorship_info.refund_amount)?;
+            }
+        }
```
joeykraut, line 277: probably simpler to just return early:
```rust
let signed_info = match assemble_sponsored_match_req.signed_sponsorship_info {
    None => return assemble_sponsored_match_req,
    Some(info) => info,
};

// Validation logic

joeykraut (approved):

auth/auth-server/src/server/handle_external_match/gas_sponsorship/refund_calculation.rs
akirillo: i have not checked yet - need to update the JS SDK first so I can test reliably. so if this is flaky due to precision the fixes will come in a follow-up PR

PR #124, opened 2025-03-14
auth-server: handle_external_match: new gas sponsorship defaults

akirillo:
This PR sets appropriate default parameters for gas sponsorship. Concretely, we have to disable in-kind gas sponsorship by default in the qutoe request endpoint for backwards compatibility with SDKs. Additionally, we deprecate the `use_gas_sponsorship` query parameter in favor of `omit_gas_sponsorship`, for more intuitive `false` defaults. This is a breaking change - un-updated clients that try to explicitly disable gas sponsorship with `use_gas_sponsorship=false` will have this parameter ignored. However, un-updated clients will enjoy gas sponsorship being enabled by default for direct matches.

### Testing
- [x] Test against testnet w/ un-updated SDK

### DIFF
--- a/auth/auth-server-api/src/lib.rs
+++ b/auth/auth-server-api/src/lib.rs
@@ -169,11 +169,8 @@ pub struct SponsoredMatchResponse {
 /// The query parameters used for gas sponsorship
 #[derive(Debug, Serialize, Deserialize)]
 pub struct GasSponsorshipQueryParams {
-    /// Whether to use gas sponsorship for the external match.
-    #[deprecated(since = "0.1.0", note = "Use `omit_gas_sponsorship` instead")]
-    pub use_gas_sponsorship: Option<bool>,
-    /// Whether to omit gas sponsorship for the external match.
-    pub omit_gas_sponsorship: Option<bool>,
+    /// Whether to disable gas sponsorship for the external match.
+    pub disable_gas_sponsorship: Option<bool>,
     /// The address to refund gas to.
     /// In the case of a native ETH refund, defaults to `tx::origin`.
     /// In the case of an in-kind refund, defaults to the receiver.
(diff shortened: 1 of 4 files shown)


auth/auth-server-api/src/lib.rs
```diff
@@ -169,16 +169,14 @@ pub struct SponsoredMatchResponse {
 /// The query parameters used for gas sponsorship
 #[derive(Debug, Serialize, Deserialize)]
 pub struct GasSponsorshipQueryParams {
-    /// Whether to use gas sponsorship for the external match.
-    /// Defaults to `true`.
-    pub use_gas_sponsorship: Option<bool>,
+    /// Whether to omit gas sponsorship for the external match.
+    pub omit_gas_sponsorship: Option<bool>,
```
joeykraut, line 173: I'd call this `disable`, `omit` would seem to imply we're "omitting information" whereas "disable" implies we're turning off some functionality

auth/auth-server-api/src/lib.rs
```diff
@@ -192,19 +190,19 @@ impl GasSponsorshipQueryParams {
             .unwrap_or(Address::ZERO)
     }
 
-    /// Get the gas sponsorship parameters, defaulting to the
-    /// server's defaults if not provided
-    pub fn get_or_default(&self) -> (bool, Address, bool) {
+    /// Get the gas sponsorship parameters, defaulting to in-kind gas
+    /// sponsorship
+    pub fn get_or_default_in_kind(&self) -> (bool, Address, bool) {
```
joeykraut, line 195: nit: the method name is probably fine as `get_or_default` right? It's not very typical for a method name to describe the default behavior

auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -218,9 +220,13 @@ impl Server {
...
         query_params: &GasSponsorshipQueryParams,
     ) -> Result<SponsoredQuoteResponse, AuthServerError> {
-        // Parse query params
-        let (sponsorship_requested, refund_address, refund_native_eth) =
-            query_params.get_or_default();
+        // Parse query params.
+        // For backwards compatibility, we do *not* enable in-kind gas sponsorship by
+        // default for quote requests.
+        // TODO: Once clients have updated, we should enable this by default.
+        let sponsorship_requested = query_params.omit_gas_sponsorship.unwrap_or(true);
```
joeykraut, line 227: shouldn't this invert the param here? I.e. if `omit_gas_sponsorship=True`, this will request sponsorship.

joeykraut (approved):

PR #129, opened 2025-03-22
funds manager server: modifications for switching execution venues

sehyunc:
### Purpose
This PR modifies the `/get-execution-quote` endpoint to essentially proxy requests to the LiFi `/quote` endpoint. This is because we don't need to any type validation of query params and/or response bodies as we can just forward requests, responses, errors. 

In a future PR, we will sign the quote for verification in the `/execute-swap` endpoint.

This requires moving the erc20 approval into the `/execute-swap` route.

### Testing
- [x] Tested fetching / executing quotes locally on mainnet env

### DIFF
--- a/funds-manager/funds-manager-api/src/types/quoters.rs
+++ b/funds-manager/funds-manager-api/src/types/quoters.rs
@@ -15,6 +15,17 @@ pub const GET_DEPOSIT_ADDRESS_ROUTE: &str = "deposit-address";
 /// The route to withdraw funds from custody
 pub const WITHDRAW_CUSTODY_ROUTE: &str = "withdraw";
 /// The route to fetch an execution quote on the quoter hot wallet
+///
+/// Expected query parameters (proxied directly to LiFi API):
+/// - fromChain: Source chain ID
+/// - toChain: Destination chain ID
+/// - fromToken: Source token address
+/// - toToken: Destination token address
+/// - fromAddress: Source wallet address
+/// - toAddress: Destination wallet address
+/// - fromAmount: Source token amount
+/// - order: Order preference for routing (e.g. 'CHEAPEST')
+/// - slippage: Slippage tolerance as a decimal (e.g. 0.0001 for 0.01%)
 pub const GET_EXECUTION_QUOTE_ROUTE: &str = "get-execution-quote";
 /// The route to execute a swap on the quoter hot wallet
 pub const EXECUTE_SWAP_ROUTE: &str = "execute-swap";
@@ -56,8 +67,6 @@ pub struct ExecutionQuote {
     /// The amount of tokens to sell
     #[serde(with = "u256_string_serialization")]
     pub sell_amount: U256,
-    /// The quoted price
-    pub price: String,
     /// The submitting address
     #[serde(with = "address_string_serialization")]
     pub from: Address,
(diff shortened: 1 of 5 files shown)


funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -80,22 +78,9 @@ pub struct ExecutionQuote {
...
-    /// The amount of tokens to sell
-    pub sell_amount: u128,
-}
-
-/// The response body for fetching a quote from the execution venue
-#[derive(Debug, Serialize, Deserialize)]
 pub struct GetExecutionQuoteResponse {
     /// The quote, directly from the execution venue
-    pub quote: ExecutionQuote,
+    pub quote: serde_json::Value,
```
joeykraut, line 83: is the type here knowable? I get that we don't currently need it, but opaquely using a `serde_json::Value` struct is probably bad form. If we don't need it, you may as well just remove the `GetExecutionQuoteResponse` type altogether.

funds-manager/funds-manager-server/src/main.rs
```diff
@@ -241,13 +241,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
         .and(with_server(server.clone()))
         .and_then(get_deposit_address_handler);
 
-    let get_execution_quote = warp::post()
+    let get_execution_quote = warp::get()
```
joeykraut, line 244: just calling this out, this PR is not backwards compatible. that's okay bc the clients are internal, but want to make sure we don't incur unexpected downtime as a result

joeykraut (approved):

funds-manager/funds-manager-server/src/execution_client/quotes.rs
```diff
@@ -1,121 +1,23 @@
...
-        let params = [
-            (BUY_TOKEN, buy.as_str()),
-            (SELL_TOKEN, sell.as_str()),
-            (SELL_AMOUNT, amount_str.as_str()),
-            (TAKER_ADDRESS, recipient.as_str()),
-        ];
+        query_params: HashMap<String, String>,
+    ) -> Result<serde_json::Value, ExecutionClientError> {
+        let params: Vec<(&str, &str)> =
+            query_params.iter().map(|(k, v)| (k.as_str(), v.as_str())).collect();
```
joeykraut, line 19: now that we just proxy the request, it'd be good to document the parameters we (or lifi) expect somewhere in the funds manager. the api create probably makes the most sense.

funds-manager/funds-manager-api/src/types/quoters.rs
sehyunc: The type is knowable, there are just ~100 fields, most of which we don't use. I could define a type with only those that we do, but personally think it's less code to maintain if we remove the response type altogether. Will move forward with this, and if there's a reason to define those types in Rust in the future happy to do so.

funds-manager/funds-manager-server/src/execution_client/quotes.rs
sehyunc: Sg, will document the subset of parameters we care about (again, there are many parameters we don't use).

funds-manager/funds-manager-api/src/types/quoters.rs
sehyunc: Actually, I want to keep the `GetExecutionQuoteResponse` struct so that in the future I can add a `signature` field to it for the signed tx data.

PR #130, opened 2025-03-25
funds-manager-api, server: generate and verify quote signature

sehyunc:
### Purpose
This PR ensures only quotes that originated from the funds manager are executed. I do so by signing the quote and including this when returning quotes to clients. This signature is then verified when clients attempt to execute a quote.

We use a separate HMAC key for signing quotes than authenticating API requests. It will also be stored in ASM.

### Testing
- [x] Tested locally against mainnet env

### DIFF
--- a/funds-manager/funds-manager-api/Cargo.toml
+++ b/funds-manager/funds-manager-api/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 
 renegade-api = { package = "external-api", workspace = true }
-renegade-common = { package = "common", workspace = true }
+renegade-common = { package = "common", workspace = true, default-features = false, features = ["hmac"] }
 
 ethers = "2"
 hex = "0.4.3"
--- a/funds-manager/funds-manager-api/Cargo.toml
+++ b/funds-manager/funds-manager-api/Cargo.toml
@@ -6,10 +6,10 @@ edition = "2021"
 [dependencies]
 
 renegade-api = { package = "external-api", workspace = true }
+renegade-common = { package = "common", workspace = true }
 
 ethers = "2"
 hex = "0.4.3"
-hmac = "0.12.1"
 http = "0.2.12"
 itertools = "0.13.0"
 serde = { version = "1.0", features = ["derive"] }
--- a/funds-manager/funds-manager-api/src/auth.rs
+++ b/funds-manager/funds-manager-api/src/auth.rs
@@ -1,10 +1,6 @@
 //! Auth helpers for the API
-use crate::types::quoters::ExecutionQuote;
-use hex;
-use hmac::{Hmac, Mac};
 use http::HeaderMap;
 use itertools::Itertools;
-use sha2::Sha256;
 
 /// The header key for the HMAC signature
 pub const X_SIGNATURE_HEADER: &str = "X-Signature";
@@ -12,29 +8,20 @@ pub const X_SIGNATURE_HEADER: &str = "X-Signature";
 /// signature
 pub const RENEGADE_HEADER_PREFIX: &str = "x-renegade-";
 
-/// Compute an hmac for the given request
-pub fn compute_hmac(
-    hmac_key: &[u8],
-    method: &str,
-    path: &str,
-    headers: &HeaderMap,
-    body: &[u8],
-) -> Vec<u8> {
-    // Construct the MAC
-    let mut mac = Hmac::<Sha256>::new_from_slice(hmac_key).expect("HMAC error");
+/// Get the bytes for the given request for HMAC computation
+pub fn get_request_bytes(method: &str, path: &str, headers: &HeaderMap, body: &[u8]) -> Vec<u8> {
+    // Build the message to sign
+    let mut message = Vec::new();
+    message.extend_from_slice(method.as_bytes());
+    message.extend_from_slice(path.as_bytes());
+    message.extend(get_header_bytes(headers));
+    message.extend_from_slice(body);
 
-    // Update with method, path, headers and body in order
-    mac.update(method.as_bytes());
-    mac.update(path.as_bytes());
-    add_headers_to_hmac(&mut mac, headers);
-    mac.update(body);
-
-    // Check the signature
-    mac.finalize().into_bytes().to_vec()
+    message
 }
 
-/// Hash headers into an HMAC
-fn add_headers_to_hmac(mac: &mut Hmac<Sha256>, headers: &HeaderMap) {
+/// Get bytes from headers for HMAC computation
+fn get_header_bytes(headers: &HeaderMap) -> Vec<u8> {
     let mut renegade_headers = headers
         .iter()
         .filter_map(|(k, v)| {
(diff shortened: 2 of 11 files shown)


funds-manager/funds-manager-api/src/auth.rs
```diff
@@ -51,3 +53,29 @@ fn add_headers_to_hmac(mac: &mut Hmac<Sha256>, headers: &HeaderMap) {
...
+        quote.sell_token_address,
+        quote.sell_amount,
+        quote.buy_amount,
+        quote.from,
+        quote.to,
+        hex::encode(&quote.data),
+        quote.value,
+        quote.gas_price,
+        quote.estimated_gas
+    );
```
joeykraut, line 72: i'd move this to a method on the execution quote type, maybe `ExecutionQuote::get_canonical_string_repr` or whatever name you think is best. 

to me, this makes sense from a code organization perspective, since its a derivable property of the quote itself

funds-manager/funds-manager-server/src/main.rs
```diff
@@ -160,6 +163,19 @@ impl Cli {
...
+        self.quote_hmac_key.as_ref().map(|key| {
+            let decoded = hex::decode(key).expect("Invalid quote HMAC key");
+            if decoded.len() != 32 {
+                panic!("Quote HMAC key must be 32 bytes long");
+            }
+            let mut array = [0u8; 32];
+            array.copy_from_slice(&decoded);
+            array
+        })
+    }
```
joeykraut, line 178: you could import the hmac key type from [`renegade-common`](https://github.com/renegade-fi/renegade/blob/2ecac82f7e11b2dfff74d6a60b692fddccc7edc5/common/src/types/hmac.rs#L18). This should be how the existing key works, but I wrote the funds manager before this common type existed. 

If you're willing to change both I think that'd be the best way to ensure we're using consistent hmac key formats and signatures across our stack.

Plus it has helpers that obviate the need to re-implement key deserialization, signing, verifying, etc.

funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -150,14 +151,26 @@ pub(crate) async fn get_execution_quote_handler(
...
-    let quote = server
+    let raw_quote = server
         .execution_client
         .get_quote(query_params)
         .await
         .map_err(|e| warp::reject::custom(ApiError::InternalError(e.to_string())))?;
 
-    let resp = GetExecutionQuoteResponse { quote };
+    let quote = ExecutionQuote::try_from(raw_quote)
+        .map_err(|e| warp::reject::custom(ApiError::InternalError(e)))?;
```
joeykraut, line 161: we should just deserialize the type as an `ExecutionQuote`, what's the reason for going through serde json's `Value` type first?

funds-manager/funds-manager-api/src/types/venue.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct Estimate {
+    pub from_amount: String,
+    pub to_amount: String,
+}
+
+impl TryFrom<serde_json::Value> for ExecutionQuote {
```
joeykraut, line 63: we should just deserialize directly to the `ExecutionQuote` type here. Implementing `TryFrom<Value> for ExecutionQuote` is exactly the point of serde!

funds-manager/funds-manager-server/src/server.rs
```diff
@@ -61,6 +61,8 @@ pub(crate) struct Server {
     pub aws_config: SdkConfig,
     /// The HMAC key for custody endpoint authentication
     pub hmac_key: Option<[u8; 32]>,
+    /// The HMAC key for signing quotes
+    pub quote_hmac_key: Option<[u8; 32]>,
```
joeykraut, line 65: don't think this should be optional. in what case would we want it to be? 

plus making it optional adds odd error handling in the route handlers to check that the key exists. I propose we just make it required; it's pretty core to the funds manager's functionality.

funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -150,14 +151,26 @@ pub(crate) async fn get_execution_quote_handler(
...
 
+    let hmac_key = match &server.quote_hmac_key {
+        Some(hmac_key) => hmac_key,
+        None => {
+            return Err(warp::reject::custom(ApiError::InternalError(
+                "Quote HMAC key not configured".to_string(),
+            )))
+        },
+    };
+    let signature = compute_quote_hmac(hmac_key, &quote);
```
joeykraut, line 171: I'd just implement a method `Server::sign_quote` which handles this -- that'll better encapsulate the behavior the server provides.

joeykraut (commented):
some ideas for improvement

funds-manager/funds-manager-api/Cargo.toml
```diff
@@ -6,10 +6,10 @@ edition = "2021"
 [dependencies]
 
 renegade-api = { package = "external-api", workspace = true }
+renegade-common = { package = "common", workspace = true }
```
joeykraut, line 9: fyi you can probably just turn on the [`hmac`](https://github.com/renegade-fi/renegade/blob/2ecac82f7e11b2dfff74d6a60b692fddccc7edc5/common/Cargo.toml#L8) feature in this crate:
```
renegade-common = { package = "common", workspace = true, default-features = false, features = ["hmac"] }
```

and avoid compiling in all the other features. It may fail to compile, if so that's on me and you can ignore, but let me know.

funds-manager/funds-manager-api/src/auth.rs
```diff
@@ -1,38 +1,27 @@
...
-    hmac_key: &[u8],
-    method: &str,
-    path: &str,
-    headers: &HeaderMap,
-    body: &[u8],
-) -> Vec<u8> {
-    // Construct the MAC
-    let mut mac = Hmac::<Sha256>::new_from_slice(hmac_key).expect("HMAC error");
+/// Get the bytes for the given request for HMAC computation
+pub fn get_request_bytes(method: &str, path: &str, headers: &HeaderMap, body: &[u8]) -> Vec<u8> {
```
joeykraut, line 12: the `external-api` crate has helpers for this if you like ([here](https://github.com/renegade-fi/renegade/blob/main/external-api/src/auth/auth_helpers.rs)). I've been meaning to switch the funds manager over but haven't done it yet. If you don't want to bother in this PR that's no worry, i'll pick this up when i have free cycles.

funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -87,11 +91,32 @@ pub struct ExecutionQuote {
...
+    }
+}
+
 /// The request body for fetching a quote from the execution venue
 #[derive(Debug, Serialize, Deserialize)]
 pub struct GetExecutionQuoteResponse {
     /// The quote, directly from the execution venue
-    pub quote: serde_json::Value,
+    pub quote: ExecutionQuote,
+    /// The HMAC signature of the quote
```
joeykraut, line 118: nit: here and below, it'd be "HMAC" not "HMAC signature", the latter is somewhat redundant and people tend to use "MAC" as the noun

joeykraut (approved):
very nice, just some small suggestions that might make your life easier

sehyunc:
> very nice, just some small suggestions that might make your life easier

i'll follow up with a separate PR for these, want to get a deploy out

PR #131, opened 2025-03-25
funds-manager: custody_client: implement withdrawal to hyperliquid

akirillo:
This PR adds a `/custody/quoters/withdraw-to-hyperliquid` endpoint, used to bridge funds into Hyperliquid. We do this from the quoter hot wallet to simplify custody management.

This requires first transferring USDC from the quoter hot wallet to the _same keypair we use for Hyperliquid, on Arbitrum_, before then transferring from that keypair to the bridge contract. This is necessary as the funds will be credited to the same address (on Hyperliquid) that sent them to the Arbitrum bridge.

### Testing
This was tested against testnet, where I confirmed that funds were received in Hyperliquid.
Note, that when testing this functionality, we need to use Hyperliquid's [dummy USDC token](https://sepolia.arbiscan.io/address/0x1baabb04529d43a73232b713c0fe471f7c7334d5).

### DIFF
--- a/funds-manager/funds-manager-api/src/types/quoters.rs
+++ b/funds-manager/funds-manager-api/src/types/quoters.rs
@@ -29,6 +29,8 @@ pub const WITHDRAW_CUSTODY_ROUTE: &str = "withdraw";
 pub const GET_EXECUTION_QUOTE_ROUTE: &str = "get-execution-quote";
 /// The route to execute a swap on the quoter hot wallet
 pub const EXECUTE_SWAP_ROUTE: &str = "execute-swap";
+/// The route to withdraw USDC to Hyperliquid from the quoter hot wallet
+pub const WITHDRAW_TO_HYPERLIQUID_ROUTE: &str = "withdraw-to-hyperliquid";
 
 // -------------
 // | Api Types |
@@ -107,4 +109,12 @@ pub struct ExecuteSwapRequest {
 pub struct ExecuteSwapResponse {
     /// The tx hash of the swap
     pub tx_hash: String,
+}
+
+/// The request body for withdrawing USDC to Hyperliquid from the quoter hot
+/// wallet
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WithdrawToHyperliquidRequest {
+    /// The amount of USDC to withdraw, in decimal format (i.e., whole units)
+    pub amount: f64,
 }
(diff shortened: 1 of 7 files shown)


funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -108,3 +110,11 @@ pub struct ExecuteSwapResponse {
     /// The tx hash of the swap
     pub tx_hash: String,
 }
+
+/// The request body for withdrawing USDC to Hyperliquid from the quoter hot
+/// wallet
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WithdrawToHyperliquidRequest {
+    /// The amount of USDC to withdraw
+    pub amount: f64,
```
joeykraut, line 119: can you note here that this is in decimal format

funds-manager/funds-manager-server/src/custody_client/withdraw.rs
```diff
@@ -124,4 +141,60 @@ impl CustodyClient {
...
         Ok(())
     }
+
+    /// Withdraw USDC to Hyperliquid from the quoter hot wallet
+    pub(crate) async fn withdraw_to_hyperliquid(
+        &self,
+        amount: f64,
+    ) -> Result<(), FundsManagerError> {
+        let hot_wallet =
+            self.get_hot_wallet_by_vault(DepositWithdrawSource::Quoter.vault_name()).await?;
```
joeykraut, line 151: probably good to implement a `get_quoter_hot_wallet` method that does exactly this. I believe we use this pattern a few times in the funds manager

funds-manager/funds-manager-server/src/custody_client/withdraw.rs
```diff
@@ -124,4 +141,60 @@ impl CustodyClient {
...
+
+        let hyperliquid_address = format!("{:#x}", hyperliquid_account.address());
+
+        let tx =
+            self.erc20_transfer(&usdc_token.addr, &hyperliquid_address, amount, hot_wallet).await?;
+
+        info!(
+            "Withdrew {amount} USDC from hot wallet to {hyperliquid_address}. Tx: {:#x}",
+            tx.transaction_hash
+        );
```
joeykraut, line 177: I would break this into two helpers that represent exactly the steps you described. This will annotate the flow better

joeykraut (approved):
some ideas to clean this up, but lgtm

PR #132, opened 2025-03-26
funds-manager: custody-client: use HL dummy USDC in testnet

akirillo:
This PR ensures that we use Hyperliquid's [dummy USDC token](https://sepolia.arbiscan.io/address/0x1baabb04529d43a73232b713c0fe471f7c7334d5) in testnet as opposed to our own USDC token in the token mapping.

### Testing
This was tested by running a funds manager locally and triggering a withdrawal to hyperliquid

### DIFF
--- a/funds-manager/funds-manager-server/src/custody_client/withdraw.rs
+++ b/funds-manager/funds-manager-server/src/custody_client/withdraw.rs
@@ -17,10 +17,12 @@ use super::{CustodyClient, DepositWithdrawSource};
 
 /// The suffix for the secret name for the Hyperliquid private key
 const HYPERLIQUID_PKEY_SECRET_SUFFIX: &str = "hyperliquid-private-key";
-/// The address of the Hyperliquid bridge on Arbitrum mainnet.
+/// The address of the Hyperliquid bridge on Arbitrum mainnet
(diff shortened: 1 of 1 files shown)


funds-manager/funds-manager-server/src/custody_client/withdraw.rs
```diff
@@ -152,8 +154,13 @@ impl CustodyClient {
         amount: f64,
     ) -> Result<(), FundsManagerError> {
         let hot_wallet = self.get_quoter_hot_wallet().await?;
-        let usdc_mint = Token::from_ticker(USDC_TICKER).get_addr();
-        let bal = self.get_erc20_balance(&usdc_mint, &hot_wallet.address).await?;
+
+        let usdc_mint = match self.chain {
+            Chain::Mainnet => &Token::from_ticker(USDC_TICKER).addr,
```
joeykraut, line 159: I added a `Token::usdc` method by the way, you could use that here

joeykraut (approved):

funds-manager/funds-manager-server/src/custody_client/withdraw.rs
akirillo: going to punt on this, doing so requires updating relayer deps => upgrading alloy version => tweaking alloy usage across funds manager & auth server

PR #139, opened 2025-04-02
funds-manager: execution_client: use provided estimated gas limit

sehyunc:
### Purpose
This PR adds and uses the gas limit field to the `ExecutionQuote` fetched from the execution venue. This fixes transactions reverting due to `out of gas`.

### Testing
- [x] Tested locally (mainnet env)

### DIFF
--- a/funds-manager/funds-manager-api/src/types/quoters.rs
+++ b/funds-manager/funds-manager-api/src/types/quoters.rs
@@ -94,13 +94,16 @@ pub struct ExecutionQuote {
     /// The estimated gas for the swap
     #[serde(with = "u256_string_serialization")]
     pub estimated_gas: U256,
+    /// The gas limit for the swap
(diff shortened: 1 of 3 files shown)


funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -94,13 +94,16 @@ pub struct ExecutionQuote {
     /// The estimated gas for the swap
     #[serde(with = "u256_string_serialization")]
     pub estimated_gas: U256,
```
sehyunc, line 96: Decided to keep this in case we need it, but we currently do not use it anywhere. Let me know if you think we should remove it for simplicity.

sehyunc:
This isn't backwards compatible so the bot server and admin dashboard will be updated at the same time.

joeykraut (approved):

PR #140, opened 2025-04-02
funds-manager: quote: validate tx receipt status

sehyunc:
### Purpose
This PR checks the transaction receipt status before returning, erroring if it reverted. This fixes an issue where reverted transactions would be logged as successful because we received the receipt, but with status = failed.

### Testing
- [x] Tested locally
- [ ] Test in testnet

### DIFF
--- a/funds-manager/funds-manager-server/src/execution_client/swap.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/swap.rs
@@ -63,10 +63,18 @@ impl ExecutionClient {
             .send_transaction(tx, None /* block */)
             .await
             .map_err(ExecutionClientError::arbitrum)?;
-        pending_tx
+        let receipt = pending_tx
             .await
             .map_err(ExecutionClientError::arbitrum)?
-            .ok_or_else(|| ExecutionClientError::arbitrum("Transaction failed"))
+            .ok_or_else(|| ExecutionClientError::arbitrum("Transaction failed"))?;
+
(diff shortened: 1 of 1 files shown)


funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -63,10 +63,18 @@ impl ExecutionClient {
...
+        let receipt = pending_tx
             .await
             .map_err(ExecutionClientError::arbitrum)?
-            .ok_or_else(|| ExecutionClientError::arbitrum("Transaction failed"))
+            .ok_or_else(|| ExecutionClientError::arbitrum("Transaction failed"))?;
+
+        let status = receipt.status.expect("status is `Some` after EIP-658");
+        if status.is_zero() {
+            let error_msg = format!("tx ({:#x}) failed with status 0", receipt.transaction_hash);
+            return Err(ExecutionClientError::arbitrum(error_msg));
```
joeykraut, line 74: can we add revert data here if available?

joeykraut (approved):

funds-manager/funds-manager-server/src/execution_client/swap.rs
sehyunc: personally think it's a little heavy handed to do tx simulation to retrieve the revert message here, especially since it's 3 clicks away once you have the hash. 
the error handling above will log revert data before the tx has been submitted. 
if we find that many txns are failing after submission, I'm happy to add this in for more helpful logs

PR #152, opened 2025-04-09
auth-server: handle_external_match: update quote request to account for sponsorship

akirillo:
This PR updates the quote request endpoint to properly respect requested exact output amounts for sponsored quotes. Namely, this means that when a quote is requested with an exact amount of the output token specified, _and_ in-kind gas sponsorship directed to the receiver is applied, we subtract the refund amount from the exact output amount before sending the quote request to the relayer.

As a result, the relayer will produce a _smaller_ quote which will bear the exact output amount requested when the refund is issued alongside it. We ensure that the exact output amount in the quote response sent to the user matches what they originally requested.

This requires us to compute a refund amount _prior_ to receiving a match result/quote from the relayer, which can be done efficiently by leveraging the new multi-price stream.

### TODO
- [ ] Update the quote assembly path (must properly revert gas sponsorship on order in signed quote, and handle updated orders)
- [ ] Update direct match path (effectively replicates this process)
- [ ] Testing

### DIFF
--- a/auth/auth-server/src/server/handle_external_match/mod.rs
+++ b/auth/auth-server/src/server/handle_external_match/mod.rs
@@ -18,7 +18,10 @@ use renegade_constants::EXTERNAL_MATCH_RELAYER_FEE;
 use tracing::{error, info, instrument, warn};
 use warp::{reject::Rejection, reply::Reply};
 
-use super::helpers::{generate_quote_uuid, overwrite_response_body};
+use super::helpers::{
+    apply_gas_sponsorship_to_exact_output_amount, generate_quote_uuid, overwrite_response_body,
+    requires_exact_output_amount_update,
+};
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -414,7 +438,10 @@ impl Server {
 
         // Compute refund amount
         let refund_amount = self
-            .get_refund_amount(&external_match_resp.match_bundle.match_result, refund_native_eth)
+            .compute_refund_amount_for_order(
+                &external_match_resp.match_bundle.match_result,
```
akirillo, line 442: this is an expected compiler error at the moment, I will update this code path in a follow-up PR

joeykraut (approved):

PR #153, opened 2025-04-10
auth-server: handle_external_match: sponsored exact output in assembly & direct match

akirillo:
This PR ensures that exact output amounts are properly respected both when assembling quotes, and executing direct matches.

Concretely, this entails ensuring that the refund amount is subtracted from the order (or updated order) being assembled / matched directly.

This is necessary to ensure the bundle served by the relayer matches a smaller order size such that the exact output amount is met post-refund, and in the case of assembly, also ensures that the quote matches what was signed by the relayer.

### Testing
Testing is deferred to the following PR, as this one resulted in a fair bit of cleanup due to generating gas sponsorship data before proxying any requests to the relayer, and I didn't want this review to snowball

### DIFF
--- a/auth/auth-server/src/server/handle_external_match/gas_sponsorship/refund_calculation.rs
+++ b/auth/auth-server/src/server/handle_external_match/gas_sponsorship/refund_calculation.rs
@@ -1,6 +1,7 @@
 //! Logic for calculating refund info for a sponsored match
 
 use alloy_primitives::U256 as AlloyU256;
+use auth_server_api::GasSponsorshipInfo;
 use bigdecimal::{BigDecimal, FromPrimitive};
 use renegade_api::http::external_match::{
     ApiExternalMatchResult, ApiExternalQuote, AtomicMatchApiBundle, ExternalOrder,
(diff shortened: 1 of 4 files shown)


auth/auth-server/src/server/handle_external_match/gas_sponsorship/refund_calculation.rs
```diff
@@ -176,3 +189,54 @@ pub(crate) fn update_match_result_with_gas_sponsorship(
...
+) {
+    match order.side {
+        OrderSide::Buy => {
+            order.exact_base_output += gas_sponsorship_info.refund_amount;
+        },
+        OrderSide::Sell => {
+            order.exact_quote_output += gas_sponsorship_info.refund_amount;
+        },
+    }
+}
```
akirillo, line 242: these methods were directly copied over from `helpers.rs`

auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -354,79 +357,70 @@ impl Server {
...
-            self.construct_sponsored_match_response(
-                external_match_resp,
-                refund_native_eth,
-                refund_address,
-                refund_amount,
-            )?
-        } else if query_params.use_gas_sponsorship.unwrap_or(false) {
-            // If the deprecated `use_gas_sponsorship` query parameter is set, it means an
+        let gas_sponsorship_info = if gas_sponsorship_info.is_none()
+            && query_params.use_gas_sponsorship.unwrap_or(false)
```
akirillo, line 365: now that we're doing automatic gas sponsorship, there's no point in keeping this backwards-compatibility logic, right? if gas sponsorship info isn't present in redis it must mean that it was either explicitly disabled when the quote was requested (only possible w/ updated SDK) or the rate limit was hit

joeykraut: makes sense to me, feel free to remove

joeykraut (approved):
clean, i like the method organization in this pr

PR #156, opened 2025-04-15
auth-server: add debug logs to identify error in calculating req quote amount

sehyunc:
### Purpose
This PR adds debug logging to track an intermittent issue where `requested_quote_amount` is 0 for Buy orders despite having non-zero `response_quote_amount` and `base_fill_ratio`.

The logging will help identify:
1. Which calculation step produces the zero value
2. Whether we're hitting the exact quote output branch
3. If the price calculation for Buy orders is correct

This will allow comparison of successful vs failed cases to pinpoint the root cause.

### Testing
- [ ] Test in testnet


joeykraut (approved):
you may also want to setup traces in the auth server so that you can easily match this to a corresponding request or other logs

sehyunc:
> you may also want to setup traces in the auth server so that you can easily match this to a corresponding request or other logs

good call, for the time being i will just use `request_id`

PR #161, opened 2025-04-17
auth-server: add on-chain-event-listener worker

sehyunc:
### Purpose
This PR adds the initial implementation of the `chain-events` listener that will listen for `NullifierSpent` events and handle post-settlement tasks such as recording metrics. It does not include metrics recording logic, that will come in a future PR. 

It also moves the create_arbitrum_client` helper into a helpers file.

### Todo
- [ ] downstream websocket fixes after Joey merges

### Testing
- [ ] Tested locally
- [ ] Test in testnet


auth/auth-server/src/main.rs
```diff
@@ -193,8 +194,29 @@ async fn main() {
...
     let system_clock = SystemClock::new().await;
 
+    // Create the arbitrum client
+    let arbitrum_client = helpers::create_arbitrum_client(
+        args.darkpool_address.clone(),
+        args.chain_id,
+        args.rpc_url.clone(),
+    )
+    .await
+    .expect("failed to create arbitrum client");
```
joeykraut, line 204: i would just have the chain events worker create the arbitrum client, since it's not used anywhere else

auth/auth-server/src/chain_events/worker.rs
```diff
@@ -0,0 +1,63 @@
...
+            })
+            .map_err(|err| OnChainEventListenerError::Setup(err.to_string()))?;
+
+        self.executor_handle = Some(join_handle);
+        Ok(())
+    }
+
+    /// Spawns a watcher thread that joins the given handle and logs its
+    /// outcome.
+    pub fn watch(mut self) {
```
joeykraut, line 50: do we need all this structure and a separate runtime? this is relevant in the relayer for relayer-specific reasons but i'm not sure we need to inherit the complexity here. I'll defer to you though, we can always simplify later if you want to experiment with this approach

joeykraut (approved):

auth/auth-server/src/main.rs
sehyunc: The `Server` uses it

auth/auth-server/src/chain_events/worker.rs
sehyunc: fair, we don't need a separate runtime for the watcher so will remove that. I like the worker pattern though, will keep it for now and possibly simplify later if it indeed only adds complexity.

PR #163, opened 2025-04-18
auth-server: add bundle store

sehyunc:
### Purpose
This PR adds a bundle store so the asynchronous `chain-events` worker can access API level details about a bundle when processing settlement. 

We maintain two in-memory maps:
-  bundle IDs -> bundle contexts
-  nullifier -> bundle IDs

Our workload prioritizes writes from the HTTP handler and only ever has one reader that is not time-sensitive, so we use a Mutex to guarantee atomic writes (over something like a RwLock). While this does mean all operations are serialized, the nature of our workload means we are okay with reads queuing up behind writes.

We remove bundle contexts on `NullifierSpent` to prevent unbounded memory growth. By indexing nullifier -> bundle_id, each cleanup holds the lock for at most O(M) (M = `MAX_CONCURRENT_TASKS`) instead of O(N).

### Todo
- [ ] Bundle IDs are derived from the nullifier and amounts parameterizing the match. This should be improved, as it allows for collisions between two bundles with the exact same {base | quote} {mint | amount}. 

### Testing
- [ ] Tested locally
- [ ] Test in testnet

### DIFF
--- a/auth/auth-server/src/main.rs
+++ b/auth/auth-server/src/main.rs
@@ -215,7 +215,7 @@ async fn main() {
     .expect("failed to create arbitrum client");
 
     // Create the shared in-memory bundle store
-    let bundle_store = Arc::new(BundleStore::new());
+    let bundle_store = BundleStore::new();
 
     // Start the on-chain event listener
     let chain_listener_config = OnChainEventListenerConfig {
--- a/auth/auth-server/src/main.rs
+++ b/auth/auth-server/src/main.rs
@@ -23,6 +23,7 @@ pub(crate) mod models;
 #[allow(missing_docs, clippy::missing_docs_in_private_items)]
 pub(crate) mod schema;
 mod server;
+mod store;
 mod telemetry;
 
 use renegade_arbitrum_client::constants::Chain;
@@ -34,6 +35,7 @@ use reqwest::StatusCode;
 use serde_json::json;
 use std::net::SocketAddr;
 use std::sync::Arc;
+use store::BundleStore;
 use thiserror::Error;
 use tracing::{error, info};
 use uuid::Uuid;
@@ -212,19 +214,24 @@ async fn main() {
     .await
     .expect("failed to create arbitrum client");
 
+    // Create the shared in-memory bundle store
+    let bundle_store = Arc::new(BundleStore::new());
+
     // Start the on-chain event listener
     let chain_listener_config = OnChainEventListenerConfig {
         websocket_addr: args.eth_websocket_addr.clone(),
         arbitrum_client: arbitrum_client.clone(),
     };
-    let mut chain_listener = OnChainEventListener::new(chain_listener_config)
+    let mut chain_listener = OnChainEventListener::new(chain_listener_config, bundle_store.clone())
         .expect("failed to build on-chain event listener");
     chain_listener.start().expect("failed to start on-chain event listener");
     chain_listener.watch();
 
     // Create the server
     let server = Arc::new(
-        Server::new(args, &system_clock, arbitrum_client).await.expect("Failed to create server"),
+        Server::new(args, &system_clock, arbitrum_client, bundle_store)
+            .await
+            .expect("Failed to create server"),
     );
 
     // --- Management Routes --- //
--- a/auth/auth-server/src/main.rs
+++ b/auth/auth-server/src/main.rs
@@ -228,11 +228,10 @@ async fn main() {
     chain_listener.watch();
 
     // Create the server
-    let server = Arc::new(
-        Server::new(args, &system_clock, arbitrum_client, bundle_store)
-            .await
-            .expect("Failed to create server"),
-    );
+    let server_inner = Server::new(args, &system_clock, arbitrum_client, bundle_store)
+        .await
+        .expect("Failed to create server");
+    let server = Arc::new(server_inner);
 
     // --- Management Routes --- //
 
--- a/auth/auth-server/src/server/handle_external_match/mod.rs
+++ b/auth/auth-server/src/server/handle_external_match/mod.rs
@@ -25,7 +25,11 @@ use warp::{reject::Rejection, reply::Reply};
 use super::helpers::{generate_quote_uuid, overwrite_response_body};
 use super::Server;
 use crate::error::AuthServerError;
-use crate::telemetry::helpers::{calculate_implied_price, record_relayer_request_500};
+use crate::store::helpers::generate_bundle_id;
+use crate::store::BundleContext;
+use crate::telemetry::helpers::{
+    calculate_implied_price, extract_nullifier_from_match_bundle, record_relayer_request_500,
+};
 use crate::telemetry::labels::{GAS_SPONSORED_METRIC_TAG, SDK_VERSION_METRIC_TAG};
 use crate::telemetry::{
     helpers::{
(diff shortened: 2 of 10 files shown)


auth/auth-server/src/store/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    by_id: HashMap<String, BundleContext>,
+    /// nullifier → queue of bundle_ids
+    /// TODO: Maybe use String instead of Nullifier
+    /// TODO: Is VecDeque the best data structure here?
+    by_null: HashMap<Nullifier, VecDeque<String>>,
+}
+
+pub struct BundleStore {
+    inner: Mutex<StoreInner>,
+}
```
joeykraut, line 38: Two things here:
1. I'd wrap `inner` in an `Arc` directly and `#[derive(Clone)]` on the type. That way you can pass around the `BundleStore` type directly, which is a bit cleaner. In effect, we're internalizing all concurrency safety to the type itself.
2. I'd use a [`RwLock`](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html) here, unless _all_ access locations are writers.

auth/auth-server/src/store/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+
+impl BundleStore {
+    pub fn new() -> Self {
+        Self { inner: Mutex::new(StoreInner { by_id: HashMap::new(), by_null: HashMap::new() }) }
```
joeykraut, line 42: Probably would be cleaner to define `StoreInner::new` for this

auth/auth-server/src/store/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// and we `yield_now()` to give the executor a chance to schedule
+    /// the writer next.  Once the lock is free, we grab it instantly.
+    pub async fn read(&self, bundle_id: &str) -> Result<Option<BundleContext>, AuthServerError> {
+        loop {
+            if let Ok(idx) = self.inner.try_lock() {
+                return Ok(idx.by_id.get(bundle_id).cloned());
+            }
+            tokio::task::yield_now().await;
+        }
+    }
```
joeykraut, line 73: As mentioned above, I'd use `RwLock` for this, it will handle the write prioritization and has really efficient optimistic spin wait behavior, before it backs off to a condition variable

auth/auth-server/src/store/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// The nullifier that was nullified as a result of the bundle being settled
+    pub nullifier: Nullifier,
+}
+
+struct StoreInner {
+    /// bundle_id → context
+    by_id: HashMap<String, BundleContext>,
+    /// nullifier → queue of bundle_ids
+    /// TODO: Maybe use String instead of Nullifier
+    /// TODO: Is VecDeque the best data structure here?
```
joeykraut, line 32: as mentioned in your comment, why not a `Vec`? We don't use the double-endedness anywhere in the implementation, a `Vec` is already perfectly fine

joeykraut (commented):
some comments on the `BundleStore` before I continue reviewing

auth/auth-server/src/store/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub gas_sponsorship_info: Option<GasSponsorshipInfo>,
+    /// Whether the bundle was sponsored
+    pub is_sponsored: bool,
+    /// The nullifier that was nullified as a result of the bundle being settled
+    pub nullifier: Nullifier,
+}
+
+struct StoreInner {
+    by_id: HashMap<String, BundleContext>,
+    by_null: HashMap<Nullifier, Vec<String>>,
```
joeykraut, line 31: nit: can you add doc comments to this struct and its fields?

auth/auth-server/src/store/helpers.rs
```diff
@@ -0,0 +1,20 @@
+use alloy_primitives::{hex, keccak256};
+use renegade_api::http::external_match::ApiExternalMatchResult;
+use renegade_circuit_types::wallet::Nullifier;
+use renegade_constants::Scalar;
+
+use crate::error::AuthServerError;
+
+/// Generates a bundle ID from the match bundle
+pub fn generate_bundle_id(
```
joeykraut, line 9: what if i request two bundles that happen to have the same amount? This could happen if i assemble the same quote twice -- seeing as the price is fixed. In this case we'd have an ID collision. I see above that we'd sort of just have to choose a bundle and make do, but might be worth explicitly adding logic in the `BundleStore` to prevent collisions, or at least ensure that correctness is maintained.

auth/auth-server/src/server/mod.rs
```diff
@@ -108,6 +109,8 @@ pub struct Server {
     /// The minimum order quote amount for which gas sponsorship is allowed,
     /// in whole units of USDC
     pub min_sponsored_order_quote_amount: f64,
+    /// The bundle store
+    pub bundle_store: Arc<BundleStore>,
```
joeykraut: we don't need an `Arc` here now that it's in `BundleStore`

auth/auth-server/src/store/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub fn new() -> Self {
+        Self { inner: Arc::new(RwLock::new(StoreInner::new())) }
+    }
+
+    pub async fn write(
+        &self,
+        bundle_id: String,
+        ctx: BundleContext,
+    ) -> Result<(), AuthServerError> {
+        let mut inner = self.inner.write().await;
```
sehyunc, line 56: Curious what level of error handling is appropriate here, should I fully check for poisoned locks, missing key, pre-lock validations, etc. or is it sufficient to let the caller handle errors? My sense is that it's ok to let caller log errors since this is mainly used for metrics where errors aren't critical.

auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -506,6 +510,26 @@ impl Server {
...
+            nullifier,
+        };
+
+        // Non-blocking write to bundle store
+        let bundle_store = self.bundle_store.clone();
+        tokio::spawn(async move {
+            if let Err(e) = bundle_store.write(bundle_id, bundle_ctx).await {
+                tracing::error!("bundle_store.write failed: {}", e);
+            }
+        });
```
joeykraut, line 531: why spawn in a separate thread? if you're worried about blocking this thread, I wouldn't be. This path is unlikely to take more than a few hundred nanoseconds up to a microsecond, so it is far too cheap to worry about.

auth/auth-server/src/main.rs
```diff
@@ -212,19 +214,24 @@ async fn main() {
...
         .expect("failed to build on-chain event listener");
     chain_listener.start().expect("failed to start on-chain event listener");
     chain_listener.watch();
 
     // Create the server
     let server = Arc::new(
-        Server::new(args, &system_clock, arbitrum_client).await.expect("Failed to create server"),
+        Server::new(args, &system_clock, arbitrum_client, bundle_store)
+            .await
+            .expect("Failed to create server"),
```
joeykraut, line 234: nit: i would split this into multiple lines, it'll look better:
```rust
let server_inner = Server::new(...).await.expect(...);
let server = Arc::new(server)
```

joeykraut (approved):

auth/auth-server/src/server/handle_external_match/mod.rs
sehyunc: I was indeed worried about adding latency, good to know

PR #166, opened 2025-04-19
auth-server: handle gas sponsorship when deriving bundle id

sehyunc:
### Purpose
This PR fixes bundle ID generation to factor in gas sponsorship, since the `chain-events` worker will derive the bundle ID from amounts net of gas sponsorship, the HTTP handler should too (unless the refund is in native ETH). This bundle ID is now used as the `request_id`.

We also write to the bundle store as early as possible to ensure post-settlement tasks can occur.

### Testing
- [x] Tested locally by running each example in `rust-sdk` and ensuring bundle contexts are successfully retrieved.

### DIFF
--- a/auth/auth-server/src/server/handle_external_match/mod.rs
+++ b/auth/auth-server/src/server/handle_external_match/mod.rs
@@ -25,10 +25,7 @@ use warp::{reject::Rejection, reply::Reply};
 use super::helpers::{generate_quote_uuid, overwrite_response_body};
 use super::Server;
 use crate::error::AuthServerError;
-use crate::store::{helpers::generate_bundle_id, BundleContext};
-use crate::telemetry::helpers::{
-    calculate_implied_price, extract_nullifier_from_match_bundle, record_relayer_request_500,
-};
+use crate::telemetry::helpers::{calculate_implied_price, record_relayer_request_500};
 use crate::telemetry::labels::{GAS_SPONSORED_METRIC_TAG, SDK_VERSION_METRIC_TAG};
 use crate::telemetry::{
     helpers::{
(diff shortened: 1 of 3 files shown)


auth/auth-server/src/server/handle_external_match/mod.rs
```diff
@@ -443,11 +452,12 @@ impl Server {
         req: &AssembleExternalMatchRequest,
         headers: &HeaderMap,
         resp: &SponsoredMatchResponse,
+        request_id: Option<String>,
```
joeykraut, line 455: any reason this is optional? I don't see `None` passed anywhere

auth/auth-server/src/server/handle_external_match/settlement.rs
```diff
@@ -0,0 +1,96 @@
...
+            key_description: key.clone(),
+            request_id: bundle_id.clone(),
+            sdk_version: get_sdk_version(headers),
+            gas_sponsorship_info: sponsored_match.gas_sponsorship_info.clone(),
+            is_sponsored: sponsored_match.is_sponsored,
+            nullifier,
+        };
+
+        // Write to bundle store
+        let bundle_store = self.bundle_store.clone();
```
joeykraut, line 41: why do we need to clone here, if you're just trying to break things across lines can we just make a reference?
```rust
let bundle_store = &self.bundle_store
```

auth/auth-server/src/server/handle_external_match/mod.rs
sehyunc: was optional previously, should no longer be optional after this change

PR #170, opened 2025-04-22
funds-manager: custody_client: init rpc_shim w/ eth_accounts impl

akirillo:
This PR introduces the `rpc_shim` module to the funds manager's `CustodyClient`, which is responsible for handling a subset of RPC method calls with opinionated validation logic. This lets us use the funds manager as a custom RPC "transport" to implement [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193)-compatible clients in libraries such as [Viem](https://viem.sh/docs/clients/transports/custom).

Our concrete use case for this is to sign `eth_signTypedData_v4` requests generated when interacting with our Hyperliquid account from a keypair custodied in Fireblocks, while applying application-specific validation logic to the incoming requests, such as: "do we really want to deposit this much into Hyperliquid?"

### Testing
- [x] The `eth_accounts` RPC method was tested by running a local funds manager, and invoking the JSON RPC request manually. I subbed out the soon-to-be-made `Hyperliquid` vault for the `Quoters` vault and confirmed that the correct address was returned.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4312,6 +4312,7 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 name = "funds-manager"
 version = "0.1.0"
 dependencies = [
+ "alloy-json-rpc",
  "alloy-sol-types 1.0.0",
  "arbitrum-client",
  "aws-config",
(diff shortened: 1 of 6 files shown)


joeykraut (approved):
very clean

PR #171, opened 2025-04-23
auth-server: add admin liquidity endpoint proxy

sehyunc:
### Purpose
This PR adds the`/order_book/depth/:mint` route to the set of proxied routes.

### Testing
- [ ] Tested locally
- [x] Test in testnet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,21 +23,21 @@ lto = true
 [workspace.dependencies]
 # === Renegade Dependencies === #
 contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
-renegade-arbitrum-client = { package = "arbitrum-client", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/add-endpoint", features = [
+renegade-arbitrum-client = { package = "arbitrum-client", git = "https://github.com/renegade-fi/renegade.git", features = [
     "rand",
 ] }
-renegade-api = { package = "external-api", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/add-endpoint", features = [
+renegade-api = { package = "external-api", git = "https://github.com/renegade-fi/renegade.git", features = [
     "auth",
 ] }
-renegade-common = { package = "common", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/add-endpoint" }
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -23,21 +23,21 @@ lto = true
 [workspace.dependencies]
 # === Renegade Dependencies === #
 contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
-renegade-arbitrum-client = { package = "arbitrum-client", git = "https://github.com/renegade-fi/renegade.git", features = [
+renegade-arbitrum-client = { package = "arbitrum-client", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/add-endpoint", features = [
```
sehyunc, line 26: Will revert this before merging.

auth/auth-server/src/main.rs
```diff
@@ -271,6 +271,22 @@ async fn main() {
...
+        .and(warp::path("depth"))
+        .and(warp::path::param::<String>())
+        .and(warp::path::full())
+        .and(warp::header::headers_cloned())
+        .and(with_server(server.clone()))
+        .and_then(
+            |mint: String,
+             full_path: warp::path::FullPath,
+             headers: warp::hyper::HeaderMap,
+             server: Arc<Server>| async move {
```
joeykraut, line 285: do we need to attach types here? the function you're calling should type these values, like in the handler above. it makes the formatting weird

auth/auth-server/src/server/order_book/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+    #[instrument(skip(self, path, headers))]
+    pub async fn handle_order_book_depth_request(
+        &self,
+        path: warp::path::FullPath,
+        headers: HeaderMap,
+        mint: String,
+    ) -> Result<impl Reply, Rejection> {
+        // Authorize the request
+        let path_str = path.as_str();
+        let key_desc = self.authorize_request(path_str, "", &headers, &[]).await?;
```
joeykraut, line 21: nit: for the literals here, it's good to inline comment the args they represent e.g.
```rust
self.authorize_request(path_str, "" /* query_str */, &headers, &[] /* body */).await?;
```

joeykraut (approved):

PR #180, opened 2025-04-30
auth-server: Use `data` field on tx

joeykraut:
### Purpose
This PR sets the `data` field on the settlement transactions rather than the `input` field to match the client configurations.

### Testing
- [x] Testing in testnet


sehyunc (approved):
nice, no sdk changes

PR #186, opened 2025-05-09
funds-manager-server: custody_client: use alloy & darkpoolclient

akirillo:
This PR upgrades the `custody_client` module of the `funds-manager-server` crate to use `alloy` and the `DarkpoolClient` appropriately. Here, we have to reintroduce `ethers` as a dependency, but we use it _only_ for deserializing `TypedData` in the `rpc_shim` module for the documented reason:
> We deliberately use the Ethers implementation of EIP-712 TypedData
rather than Alloy's, because Alloy will fail to deserialize TypedData which
contains type identifiers that are not valid Solidity types.
For example, a Hyperliquid withdrawal action contains a field with the type
identifier `HyperliquidTransaction:Withdraw`. Due to the usage of the `:`
character, Alloy will fail to deserialize the TypedData.

With this, the `funds-manager-server` is able to build.

### Testing
- [x] Test local funds manager using local admin panel, asserting that blockchain interactions succeed (parsing private keys/addresses, calling contracts, submitting transactions)


funds-manager/funds-manager-server/src/custody_client/gas_sponsor.rs
```diff
@@ -209,39 +210,35 @@ impl CustodyClient {
...
+            .get_block(BlockId::latest())
             .await
             .map_err(FundsManagerError::arbitrum)?
             .ok_or(FundsManagerError::arbitrum("No latest block found".to_string()))?;
 
         let latest_basefee = latest_block
+            .header
             .base_fee_per_gas
-            .ok_or(FundsManagerError::arbitrum("No basefee found".to_string()))?;
+            .ok_or(FundsManagerError::arbitrum("No basefee found".to_string()))?
```
joeykraut, line 230: want to change this error type from `arbitrum` to `on_chain` or `blockchain`?

funds-manager/funds-manager-server/src/custody_client/mod.rs
```diff
@@ -238,19 +251,22 @@ impl CustodyClient {
 
     // --- Arbitrum JSON RPC --- //
 
-    /// Get a JSON RPC provider for the given RPC url
-    pub fn get_rpc_provider(&self) -> Result<Provider<Http>, FundsManagerError> {
-        Provider::<Http>::try_from(&self.arbitrum_rpc_url)
-            .map_err(err_str!(FundsManagerError::Arbitrum))
+    /// Get an instance of a signer with the http provider attached
+    fn get_signer(&self, wallet: PrivateKeySigner) -> DynProvider {
```
joeykraut, line 255: nit: `get_provider` is probably still an appropriate name here,  the signer is input

joeykraut (approved):

funds-manager/funds-manager-server/src/custody_client/mod.rs
akirillo: hmm, the thing is we now store a basic "provider" on the struct directly, which is used for non-signing blockchain interaction. I want this method to signify that it returns a purposefully different `DynProvider`. plus, this naming pattern was already established in the `ExecutionClient`, and I felt it makes sense there

PR #191, opened 2025-05-10
price-reporter: multi-chain token remaps

sehyunc:
### Purpose
This PR allows multiple token remaps to be configured in the price reporter.

### Testing
- [x] Tested locally
- [x] Test in testnet by loading both mainnet and testnet mappings and using frontend to make sure all prices streamed (from two difference environments)


Cargo.toml
```diff
@@ -23,19 +23,19 @@ lto = true
...
+renegade-common = { package = "common", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
+renegade-config = { package = "config", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
+renegade-constants = { package = "constants", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
+renegade-circuits = { package = "circuits", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
+renegade-circuit-types = { package = "circuit-types", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
+renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
+renegade-util = { package = "util", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap", features = ["metered-channels"] }
+renegade-price-reporter = { package = "price-reporter", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
+renegade-system-clock = { package = "system-clock", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap" }
```
sehyunc, line 38: TODO: update once available

price-reporter/src/utils.rs
```diff
@@ -268,9 +274,13 @@ pub fn requires_quote_conversion(exchange: &Exchange) -> bool {
...
     let exchange = Exchange::from_str(parts[0]).map_err(err_str!(ServerError::InvalidPairInfo))?;
-    let base = Token::from_addr(parts[1]);
-    let quote =
-        if exchange == Exchange::Renegade { Token::usdc() } else { Token::from_addr(parts[2]) };
+    let base = Token::from_addr_without_chain(parts[1]);
+    let chain = base.get_chain();
+    let quote = if exchange == Exchange::Renegade {
+        Token::from_ticker_on_chain(USDC_TICKER, chain)
+    } else {
+        Token::from_addr_on_chain(parts[2], chain)
```
sehyunc, line 284: continues to require price topics to be constructed with tokens on the same chain

price-reporter/src/main.rs
```diff
@@ -98,22 +103,28 @@ pub fn init_default_price_streams(
...
         let supported_exchanges: Vec<Exchange> = get_supported_exchanges(&base_token, config)
             .difference(&disabled_exchanges_set)
             .copied()
             .collect();
 
+        let chain = base_token.get_chain();
+
         for exchange in supported_exchanges {
-            let quote_token = default_exchange_stable(&exchange);
+            let quote_token = default_exchange_stable_on_chain(&exchange, chain);
```
joeykraut, line 127: None of this feels like it requires the chain. IMO we should basically:
- Fetch all distinct tickers
- Compute the default stablecoin on an exchange -- which is chain-independent
- Stream prices for each default ticker.

We can index price streams by ticker if we like to make things simpler, and pass tickers to `init_price_stream`, but the current setup only seems to rely on the chain for path-dependent reasons.

joeykraut (commented):
I would suggest we disintermediate the `Token` struct in this implementation. We really only need addresses at the API layer, and not so much at the price stream layer.

price-reporter/src/main.rs
sehyunc: Disregarding as per discussion

joeykraut (approved):

PR #192, opened 2025-05-11
price-reporter: use multiple remaps

sehyunc:
### Purpose
This PR allows specifying multiple chains when spinning up the price reporter. It removes the remap file path arg, opting instead to retrieve remaps from the default remote location.

### Testing
- [x] Tested locally
- [ ] Test in testnet

### DIFF
--- a/price-reporter/src/utils.rs
+++ b/price-reporter/src/utils.rs
@@ -8,6 +8,7 @@ use futures_util::StreamExt;
 use futures_util::{stream::SplitSink, Stream};
 use itertools::Itertools;
 use matchit::Router;
+use renegade_common::types::token::read_token_remaps;
 use renegade_common::types::{
     chain::Chain,
     exchange::Exchange,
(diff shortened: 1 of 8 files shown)


sehyunc:
Is it useful to be able to specify multiple token remap paths, or a path at all?

joeykraut:
> Is it useful to be able to specify multiple token remap paths, or a path at all?

definitely one path, it's useful for local testing.

price-reporter/src/utils.rs
```diff
@@ -193,10 +191,8 @@ pub struct PriceReporterConfig {
     pub ws_port: u16,
     /// The port on which the server listens for incoming http requests
     pub http_port: u16,
-    /// The path to the token remap file
-    pub token_remap_path: Option<String>,
-    /// The chain to use for token remapping
-    pub remap_chain: Chain,
+    /// The chains to use for token remapping
+    pub remap_chains: Vec<Chain>,
```
joeykraut, line 195: I feel like this should just be called `chains`; as it is more abstractly the chains that the price reporter supports, wdyt?

joeykraut (approved):

PR #195, opened 2025-05-13
funds-manager-server: cli: create chain-specific clients from config file

akirillo:
This PR updates the config pattern for the funds manager to parse chain-specific configuration parameters from a JSON file. This file can either be fetched from S3 at runtime, or provided locally (for easier local development).

Using these chain-specific configurations, we instantiate the various funds manager clients separately per-chain.

Basically all of the logic for these clients is chain-specific, and any chain-agnostic logic is cheap to replicate, making this a significantly simpler approach than splitting these clients into chain-agnostic and chain-specific constituents.

For example, while the `CustodyClient` does interact with a single, global DB, 1) pretty much every table will have a `chain` column along which to segment interaction, and 2) we use a DB pool anyway, so there is little overhead in having multiple `CustodyClient`s hold their own connections to the DB.

The server holds a mapping from `Chain -> ChainClients`, which we will index into in the API layer using a new `/:chain/` path parameter on each of the routes.

### TODO
- [ ] Access appropriate client in API according to `:chain` route param
- [ ] Test on Arbitrum Sepolia

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,19 +23,19 @@ lto = true
 [workspace.dependencies]
 # === Renegade Dependencies === #
 contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
-renegade-darkpool-client = { package = "darkpool-client", git = "https://github.com/renegade-fi/renegade.git"}
-renegade-api = { package = "external-api", git = "https://github.com/renegade-fi/renegade.git", features = [
+renegade-darkpool-client = { package = "darkpool-client", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap"}
+renegade-api = { package = "external-api", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap", features = [
     "auth",
 ] }
-renegade-common = { package = "common", git = "https://github.com/renegade-fi/renegade.git" }
-renegade-config = { package = "config", git = "https://github.com/renegade-fi/renegade.git" }
-renegade-constants = { package = "constants", git = "https://github.com/renegade-fi/renegade.git" }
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -23,19 +23,19 @@ lto = true
 [workspace.dependencies]
 # === Renegade Dependencies === #
 contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
-renegade-darkpool-client = { package = "darkpool-client", git = "https://github.com/renegade-fi/renegade.git"}
-renegade-api = { package = "external-api", git = "https://github.com/renegade-fi/renegade.git", features = [
+renegade-darkpool-client = { package = "darkpool-client", git = "https://github.com/renegade-fi/renegade.git", branch = "sehyun/chain-remap"}
```
akirillo, line 26: This is a temporary change to ingest the `PartialEq, Eq, Hash` derives on the `Chain` enum introduced in the associated PR

funds-manager/funds-manager-server/src/cli.rs
```diff
@@ -0,0 +1,275 @@
...
+            execution_client,
+            metrics_recorder,
+            fee_indexer,
+        })
+    }
+}
+
+/// Chain-specific clients used by the funds manager, parsed from a
+/// configuration object
+// TODO: Do I need to wrap these in `Arc`s?
```
joeykraut, line 259: is this todo still relevant?

joeykraut (approved):

funds-manager/funds-manager-server/src/cli.rs
akirillo: no, i've removed it in the next PR

PR #196, opened 2025-05-13
funds-manager-server: handlers: include chain path param where relevant

akirillo:
This PR has the funds manager's routes/handlers accept a `Chain` path parameter where relevant. This is in expectation of the `Chain` enum specifying `ArbitrumSepolia, ArbitrumOne, BaseSepolia, BaseMainnet` more directly than just `Mainnet, Testnet`. For this reason we also introduce an `Environment` enum that the server must be configured with, which will be used in e.g. the Hyperliquid functionality to select `Chain::ArbitrumSepolia, Chain::ArbitrumOne` as appropriate.

### DIFF
--- a/funds-manager/funds-manager-server/src/cli.rs
+++ b/funds-manager/funds-manager-server/src/cli.rs
@@ -5,7 +5,7 @@ use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
 use alloy::signers::local::PrivateKeySigner;
 use alloy_primitives::Address;
 use aws_config::SdkConfig;
-use clap::Parser;
+use clap::{Parser, ValueEnum};
 use renegade_circuit_types::elgamal::DecryptionKey;
 use renegade_common::types::{chain::Chain, hmac::HmacKey};
 use renegade_darkpool_client::{client::DarkpoolClientConfig, DarkpoolClient};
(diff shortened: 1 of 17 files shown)


funds-manager/funds-manager-server/src/cli.rs
```diff
@@ -34,8 +34,15 @@ const BLOCK_POLLING_INTERVAL: Duration = Duration::from_millis(100);
...
-type ChainClientsMap = HashMap<Chain, ChainClients>;
+
+/// The (chain-agnostic) environment the server is running in.
+#[derive(Clone, ValueEnum)]
+pub enum Environment {
+    /// The production environment
+    Production,
+    /// The development environment
+    Development,
+}
```
joeykraut, line 45: we can probably call these mainnet and testnet now that those names are freed up. It's more consistent with naming elsewhere in our infra.

funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -198,20 +219,30 @@ pub(crate) async fn withdraw_to_hyperliquid_handler(
     req: WithdrawToHyperliquidRequest,
     server: Arc<Server>,
 ) -> Result<Json, warp::Rejection> {
+    // TODO: Separate out chain-agnostic hedging client from custody client
+    let chain = match server.environment {
+        Environment::Production => Chain::Mainnet,
+        Environment::Development => Chain::Testnet,
+    };
```
joeykraut, line 226: looks like this is using an outdated `Chain` struct, is that just because a later PR updates it?

joeykraut (approved):

funds-manager/funds-manager-server/src/cli.rs
akirillo: PR #198 addresses both of your comments

joeykraut: nice

PR #201, opened 2025-05-15
price-reporter: init default tickers

sehyunc:
### Purpose
This PR initializes prices streams for default tickers without using `Token`, for simplicity.

### Testing
- [x] Tested locally
- [ ] Test in testnet

### DIFF
--- a/price-reporter/src/main.rs
+++ b/price-reporter/src/main.rs
@@ -7,24 +7,21 @@
 #![deny(clippy::needless_pass_by_value)]
 #![deny(clippy::needless_pass_by_ref_mut)]
 
-use std::{
-    collections::{HashMap, HashSet},
-    net::SocketAddr,
-};
+use std::{collections::HashSet, net::SocketAddr};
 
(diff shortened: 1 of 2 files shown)


price-reporter/src/main.rs
```diff
@@ -142,40 +133,40 @@ pub fn init_default_price_streams(
...
+        .map(|exchanges| exchanges.keys().copied().collect())
+        .unwrap_or_default();
+
     if !config.coinbase_configured() {
         supported_exchanges.remove(&Exchange::Coinbase);
     }
-    if !config.uniswap_v3_configured() {
-        supported_exchanges.remove(&Exchange::UniswapV3);
+    if config.uniswap_v3_configured() {
+        supported_exchanges.insert(Exchange::UniswapV3);
```
joeykraut, line 169: i may be wrong about this but won't uniswap show up in the exchange support map? so we want the action to be explicitly remove rather than explicitly add?

joeykraut (approved):

PR #205, opened 2025-05-15
funds-manager-server: use chain-aware token methods

akirillo:
This PR ensures that we invoke chain-specific static methods on the `Token` struct, now that the funds manager is initializing multiple token remaps.

Unfortunately, all of these code paths are only hit on a mainnet deployment: either because they are part of the execution quote generation / settlement API, or in the case of withdrawals to hyperliquid, we only construct the USDC token on mainnet (HL has its own testnet USDC that we hardcode in the testnet case).

As such, testing will be deferred until a mainnet deployment is out. Still, I have confirmed that a testnet funds manager can run when configured with both arbitrum-sepolia and base-sepolia configs.

### DIFF
--- a/funds-manager/funds-manager-api/src/types/quoters.rs
+++ b/funds-manager/funds-manager-api/src/types/quoters.rs
@@ -1,6 +1,9 @@
 //! API types for quoter management
 use alloy_primitives::{hex, Address, Bytes, U256};
-use renegade_common::types::token::{Token, USDC_TICKER};
+use renegade_common::types::{
+    chain::Chain,
+    token::{Token, USDC_TICKER},
+};
 use serde::{Deserialize, Serialize};
 
 use crate::{serialization::u256_string_serialization, u256_try_into_u128};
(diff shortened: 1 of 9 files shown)


funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -132,85 +135,85 @@ impl ExecutionQuote {
     }
 
     /// Get the buy amount as a decimal-corrected string
-    pub fn get_decimal_corrected_buy_amount(&self) -> Result<f64, String> {
+    pub fn get_decimal_corrected_buy_amount(&self, chain: Chain) -> Result<f64, String> {
```
sehyunc, line 138: While this does work, I feel it would be cleaner to specify a chain on the ExecutionQuote struct and access using `self.chain`. You could set it in the chain-aware `get_execution_quote_handler` or derive it from the LI.FI quote which contains a chain ID as per the [docs](https://apidocs.li.fi/reference/get_v1-quote).

sehyunc (approved):

funds-manager/funds-manager-api/src/types/quoters.rs
akirillo: perhaps i'll try the latter approach here. i didn't want to include a `chain` on the quote itself because then we have to update SDKs to expect that when serializing/deserializing a quote, and this field is only needed by the funds manager. i could `#[serde(skip)]` it but then we have to provide some default chain to use when deserializing, which is undefined

sehyunc: yeah the latter approach would require `chain` to be added to the ExecutionQuote class. i believe that would be the extent of it

sehyunc (approved):
lgtm, only thing is it feels weird to have an internal version of ExecutionQuote when we can just add `chain` to the ExecutionQuote in the internal sdk and obviate the need for `AugmentedExecutionQuote`. i'll defer to you on whether or not to keep it

akirillo:
> lgtm, only thing is it feels weird to have an internal version of ExecutionQuote when we can just add `chain` to the ExecutionQuote in the internal sdk and obviate the need for `AugmentedExecutionQuote`. i'll defer to you on whether or not to keep it

It's not needed in the internal SDK, and there's some hoops to jump through, such as converting a `LiFi` quote to a chain-aware `ExecutionQuote`, and backwards-compatibility to decouple this deploy w/ a bot server deploy

PR #207, opened 2025-05-16
price-reporter: fix multi-chain stream init

sehyunc:
### Purpose
This PR fixes bugs found while testing loading multiple token remaps:
- cannot rely on `default_chain()` since there are multiple
- improper usage of `derivative` crate leading to duplicate price streams 
- couldn't find tokens because we weren't lowecasing

### Testing
- [x] Tested loading arbitrum-sepolia, base-sepolia and fetching prices using WETH address on both
- [x] Test in testnet using script to make sure we get prices for all tokens (arbitrum + base)

### DIFF
--- a/price-reporter/src/main.rs
+++ b/price-reporter/src/main.rs
@@ -14,15 +14,13 @@ use http_server::HttpServer;
 use pair_info::PairInfo;
 use renegade_common::types::{
     exchange::Exchange,
-    token::{default_exchange_stable, read_exchange_support, USDC_TICKER, USDT_TICKER, USD_TICKER},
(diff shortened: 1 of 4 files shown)


price-reporter/src/main.rs
```diff
@@ -102,29 +100,18 @@ pub fn init_default_price_streams(
 
     let disabled_exchanges_set: HashSet<Exchange> = disabled_exchanges.into_iter().collect();
 
-    // Get distinct tickers from the token remap(s)
-    let distinct_tickers = get_all_distinct_tickers();
-    for base_ticker in distinct_tickers {
+    for base_token in get_all_tokens() {
```
sehyunc, line 103: We now rely on deduplication provided by the PairInfo struct

joeykraut (approved):

PR #209, opened 2025-05-17
price-reporter: prevent panic on ticker unwrap

sehyunc:
### Purpose
This PR avoids a panic given a price topic made of addresses of tokens on two different chains.

### Testing
- [x] Tested locally that topic with two different chain addresses does not panic the price reporter

### DIFF
--- a/price-reporter/src/pair_info.rs
+++ b/price-reporter/src/pair_info.rs
@@ -75,12 +75,14 @@ impl PairInfo {
             Token::from_addr_on_chain(parts[2], chain)
         };
 
-        Ok(Self::new(
-            exchange,
-            base.get_ticker().unwrap(),
-            quote.get_ticker().unwrap(),
-            Some(chain),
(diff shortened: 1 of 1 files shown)


price-reporter/src/pair_info.rs
```diff
@@ -75,12 +75,20 @@ impl PairInfo {
...
+                "{} and {} are not on the same chain",
+                base, quote
+            ))
+        })?;
+        let quote_ticker = quote.get_ticker().ok_or_else(|| {
+            ServerError::InvalidPairInfo(format!(
+                "{} and {} are not on the same chain",
+                base, quote
+            ))
+        })?;
```
akirillo, line 89: I would just make the error message `unable to get ticker for {}`. This could also happen if the token mapping is misconfigured or we request a price for some base/quote not in the mapping

akirillo (approved):

PR #210, opened 2025-05-17
funds-manager-server: use price reporter client

akirillo:
This PR updates the funds manager to use the price reporter client instead of querying the relayer for prices.

### Testing
- [x] Run the funds manager, confirm that prices are streamed successfully, redeem fees to test getting a price from the client


joeykraut (approved):
you've got a clippy lint issue, but otherwise lgtm

PR #217, opened 2025-05-19
price-reporter: allow same token on multiple remaps to be found

sehyunc:
### Purpose
This PR ensures that when a price is requested of a token that exists on multiple token lists, we find the right pair. 

Previously we used the `Token` on the default chain which would error when a pair that existed on a non-default chain was requested. 

### Testing
- [x] Tested locally that renegade topic of base mint that exists on two maps with unique quote returns correct subscription lit
- [x] Test in testnet

### DIFF
--- a/price-reporter/src/utils/pair_info.rs
+++ b/price-reporter/src/utils/pair_info.rs
@@ -44,18 +44,32 @@ impl PairInfo {
         Self { exchange, base, quote, chain: chain.unwrap_or(default_chain()) }
     }
 
+    /// Create a new pair info from two tokens
+    pub fn new_from_tokens(
+        exchange: Exchange,
+        base: &Token,
+        quote: &Token,
+    ) -> Result<Self, ServerError> {
+        assert_eq!(base.get_chain(), quote.get_chain());
+        let base_ticker = base.get_ticker().ok_or_else(|| {
+            ServerError::InvalidPairInfo(format!("unable to get ticker for {}", base))
+        })?;
+        let quote_ticker = quote.get_ticker().ok_or_else(|| {
+            ServerError::InvalidPairInfo(format!("unable to get ticker for {}", quote))
+        })?;
+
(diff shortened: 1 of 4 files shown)


price-reporter/src/utils/pair_info.rs
```diff
@@ -46,24 +46,26 @@ impl PairInfo {
 
     /// Create a new pair info with the default stable token of the given
     /// exchange
-    pub fn new_default_stable(exchange: Exchange, base_mint: &str) -> Self {
-        let (base, chain) = get_token_and_chain(base_mint).unwrap();
-        let quote_token = default_exchange_stable(&exchange, chain);
+    pub fn new_default_stable(exchange: Exchange, base_mint: &str) -> Result<Self, ServerError> {
+        let base_token = get_token_and_chain(base_mint).into_iter().next().ok_or_else(|| {
```
joeykraut, line 50: you use this pattern in a couple places wherein you take the first token from the candidate set. Maybe define a helper? Something like `get_token_and_chain_default`?

price-reporter/src/utils/pair_info.rs
```diff
@@ -85,25 +87,60 @@ impl PairInfo {
...
+        let quote_mint = parts[2];
+        let quote = Token::from_addr_on_chain(quote_mint, chain);
         let quote_ticker = quote.get_ticker().ok_or_else(|| {
             ServerError::InvalidPairInfo(format!("unable to get ticker for {}", quote))
         })?;
 
         Ok(Self::new(exchange, base_ticker, quote_ticker, Some(chain)))
     }
 
+    /// Parse a Renegade topic
```
joeykraut, line 112: nit: Will you add a doc comment explaining that renegade topics may be `renegade-<base>` or `renegade-<base>-<usdc>`?

joeykraut (approved):

PR #218, opened 2025-05-19
auth-server: server: external-match: Use contexts in quote handler

joeykraut:
### Purpose
This PR makes the following changes to the endpoint handlers in the external match flow:
- Moves each handler to its own file for clarity
- Defines a `RequestContext` and `ResponseContext` type that will be used for pre and post-processing of the request.
- Uses this type throughout the quote endpoint's post-processing metrics.

### Todo
- Use this request context type for other external match endpoints

### DIFF
--- a/auth/auth-server/src/server/api_handlers/external_match/quote.rs
+++ b/auth/auth-server/src/server/api_handlers/external_match/quote.rs
@@ -1,0 +1,80 @@
+//! Quote endpoint handler
+
+use bytes::Bytes;
+use http::{Method, StatusCode};
+use tracing::{error, instrument, warn};
+use warp::{reject::Rejection, reply::Reply};
+
+use crate::{
+    error::AuthServerError,
+    http_utils::overwrite_response_body,
+    server::{api_handlers::log_unsuccessful_relayer_request, Server},
+    telemetry::helpers::record_relayer_request_500,
+};
+
+impl Server {
+    /// Handle an external quote request
+    #[instrument(skip(self, path, headers, body))]
+    pub async fn handle_quote_request(
+        &self,
+        path: warp::path::FullPath,
+        headers: warp::hyper::HeaderMap,
+        body: Bytes,
+        query_str: String,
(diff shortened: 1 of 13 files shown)


auth/auth-server/src/server/rate_limiter/gas_sponsorship_rate_limiter.rs
```diff
@@ -58,9 +58,9 @@ impl GasSponsorshipRateLimiter {
     }
 
     /// Check if the given user's bucket has a non-zero remaining value.
-    pub async fn has_remaining_value(&self, user_id: String) -> bool {
+    pub async fn has_remaining_value(&self, user_id: &str) -> bool {
```
sehyunc, line 61: Is your rationale for String -> &str to only take ownership at the latest possible step i.e. reading from the HashMap (where you have to take the key by value)?

auth/auth-server/src/server/api_handlers/external_match/quote.rs
```diff
@@ -0,0 +1,307 @@
...
+        // 1. Run the pre-request subroutines
+        let mut ctx = self.preprocess_request(path, headers, body, query_str).await?;
+        self.pre_request(&mut ctx).await?;
+
+        // 2. Proxy the request to the relayer
+        let (raw_resp, ctx) = self.forward_request(ctx).await?;
+
+        // 3. Run the post-request subroutines
+        let resp = self.post_request(raw_resp, ctx)?;
+        Ok(resp)
```
sehyunc, line 85: nice cook

sehyunc (approved):

auth/auth-server/src/server/rate_limiter/gas_sponsorship_rate_limiter.rs
joeykraut: more that it makes the interfaces simpler to limit the amount of cloning that goes on in the callers. practically speaking the compiler is pretty good at eliding clones so I tend not to worry about them until they present a clear bottleneck

PR #219, opened 2025-05-20
price-reporter-client: use canonical exchange

sehyunc:
### Purpose
This PR modifies the price reporter client that the auth server and funds manager use to use the renegade topic when streaming prices, ensuring they retrieve the authoritative price.

### Testing
- [ ] Tested locally
- [ ] Test in testnet


akirillo (approved):
very satisfying, nice - just getting a clippy lint about an unused import

PR #233, opened 2025-05-29
funds-manager: execution-client: Use LiFi API key

joeykraut:
### Purpose
This PR adds the LiFi API key to all execution requests when configured. The requests are valid without an API key, but [rate limits]() are 120x higher with a key configured. 

This is done to prevent the rate limiting issues we ran into when going x-chain.

### Testing
- [ ] Test in mainnet

### DIFF
--- a/funds-manager/Dockerfile
+++ b/funds-manager/Dockerfile
@@ -54,5 +54,7 @@ RUN apt-get update && \
 # Copy the binary from the build stage
 COPY --from=builder /build/target/release/funds-manager /bin/funds-manager
 
+# Set log filtering
+ENV RUST_LOG="funds_manager=info,fireblocks_sdk=off,warp=warn"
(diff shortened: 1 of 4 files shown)


funds-manager/Dockerfile
```diff
@@ -54,5 +54,7 @@ RUN apt-get update && \
 # Copy the binary from the build stage
 COPY --from=builder /build/target/release/funds-manager /bin/funds-manager
 
+# Set log filtering
+ENV RUST_LOG="funds_manager=info,fireblocks_sdk=off,warp=warn"
```
akirillo, line 58: thank you 😭

akirillo (approved):

funds-manager/Dockerfile
joeykraut: it needed to be done

PR #235, opened 2025-05-29
funds-manager: custody-client: Use solidity `receive` selector

joeykraut:
### Purpose
This PR uses the solidity `receive` selector for transferring ETH to the gas sponsor contract on Base rather than the stylus implemented `receiveEth` method.

### Testing
- [ ] Testing in mainnet


akirillo (approved):
good catch

PR #239, opened 2025-05-31
auth-server: chain-events: abis: Parse external matches from calldata

joeykraut:
### Purpose
This PR adds helpers for parsing external matches (malleable or standard) from darkpool calldata. This allows the chain events listener to record settlement metrics for malleable match bundles.

### Testing
- [x] Test in testnet

### DIFF
--- a/auth/auth-server/src/chain_events/listener.rs
+++ b/auth/auth-server/src/chain_events/listener.rs
@@ -5,7 +5,7 @@ use std::{sync::Arc, thread::JoinHandle};
 
 use alloy::{
     providers::{DynProvider, Provider, ProviderBuilder, WsConnect},
-    rpc::types::Filter,
+    rpc::types::{trace::geth::CallFrame, Filter},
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/chain_events/listener.rs
```diff
@@ -5,7 +5,7 @@ use std::{sync::Arc, thread::JoinHandle};
 
 use alloy::{
     providers::{DynProvider, Provider, ProviderBuilder, WsConnect},
-    rpc::types::Filter,
+    rpc::types::{trace::geth::CallFrame, Filter},
     sol_types::SolEvent,
```
sehyunc, line 9: will clippy errors be fixed in a future PR?

joeykraut: i believe those are because the CI has misconfigured feature flags now that we use cross-chain feature flagging.

PR #243, opened 2025-06-05
renegade-solver: Dockerfile: Add dockerfile

joeykraut:
### Purpose
This PR adds a `Dockerfile` to the `renegade-solver` package, and updates the deploy script to include this service.

I also added fuzzy matching to the deploy script.

### Testing
- [x] Deployed to Base mainnet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3550,6 +3550,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "658bce805d770f407bc62102fca7c2c64ceef2fbcb2b8bd19d2765ce093980de"
 dependencies = [
  "console",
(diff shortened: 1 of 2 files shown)


deploy/src/main.rs
```diff
@@ -44,18 +44,19 @@ enum Commands {
     List,
 }
 
-// --------------
-// | Helpers    |
```
akirillo, line 48: thank you.

akirillo (approved):

deploy/src/main.rs
joeykraut: the models can write a deploy script in 10 seconds but can't count the number of dashes smh

PR #244, opened 2025-06-05
renegade-solver: uniswapx: solve: Check if order is servicable

joeykraut:
### Purpose
This PR checks if each order is serviceable by fetching supported tokens from the relayer at startup, caching them, and checking each order's tokens against them.

There are a number of improvements and edge cases documented in the todos that need to be taken care of, but these are deferred to a follow-up.

### Testing
- [x] Tested locally


renegade-solver/src/uniswapx/api_interaction.rs
```diff
@@ -0,0 +1,70 @@
+//! Code for interacting with the UniswapX API
```
joeykraut, line 1: broken out from existing code in `mod.rs`

renegade-solver/src/uniswapx/solve.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use crate::{
+    error::SolverResult,
+    uniswapx::{api_types::OrderEntity, UniswapXSolver},
+};
+
+impl UniswapXSolver {
+    /// Solve a set of orders and submit solutions to the reactor
+    pub(crate) async fn solve_order(&self, order: OrderEntity) -> SolverResult<()> {
+        // If the order is not serviceable, print it and return
```
akirillo, line 13: nit: we don't actually print the order lol, which I think is good - let's just remove that part of the comment

akirillo (approved):

renegade-solver/src/uniswapx/solve.rs
joeykraut: oops yeah I was printing locally to test which orders were filtered out, left in the comment. Nice catch

PR #250, opened 2025-06-06
funds-manager: execution_client: recursively invoke half-sized swaps

akirillo:
This PR upgrades the `swap-immediate` route to recursively attempt half-sized swaps, if a given swap failed. If any sub-swaps fail, the route handler will not error. Rather, failures are logged and any other successful swap results are collected, effectively allowing for partial filling of the total requested swap amount.

This requires us to construct the query parameters for calls to the LiFi API natively, for which we introduce the `serde_qs` dependency as it can handle the list-type query parameters using repeated key names, as expected by the LiFi API.

We remove the backwards compatibility routes from the multi-chain migration as all clients of the funds manager are migrated.

### Testing
- [ ] Test against mainnet using local funds manager & updated TS SDK

### DIFF
--- a/funds-manager/funds-manager-server/src/execution_client/swap.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/swap.rs
@@ -3,17 +3,24 @@
 use alloy::{
     eips::BlockId,
     network::TransactionBuilder,
-    providers::Provider,
+    providers::{DynProvider, Provider},
     rpc::types::{TransactionReceipt, TransactionRequest},
     signers::local::PrivateKeySigner,
 };
 use alloy_primitives::{Address, U256};
-use funds_manager_api::{quoters::ExecutionQuote, u256_try_into_u64};
-use tracing::info;
+use funds_manager_api::{
+    quoters::{AugmentedExecutionQuote, ExecutionQuote, LiFiQuoteParams},
+    u256_try_into_u64,
+};
+use renegade_common::types::chain::Chain;
+use tracing::{error, info, warn};
 
 use crate::helpers::IERC20;
 
 use super::{error::ExecutionClientError, ExecutionClient};
+
+/// The minimum amount of USDC that will be attempted to be swapped recursively
+const MIN_SWAP_QUOTE_AMOUNT: f64 = 10.0; // 10 USDC
 
 impl ExecutionClient {
     /// Execute a quoted swap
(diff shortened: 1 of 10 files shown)


funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -41,6 +55,28 @@ impl ExecutionClient {
...
+        // Send the transaction
+        let pending_tx =
+            client.send_transaction(tx).await.map_err(ExecutionClientError::arbitrum)?;
+
+        let receipt = pending_tx.get_receipt().await.map_err(ExecutionClientError::arbitrum)?;
+
+        if !receipt.status() {
+            let error_msg = format!("tx ({:#x}) failed", receipt.transaction_hash);
+            return Err(ExecutionClientError::arbitrum(error_msg));
+        }
```
joeykraut, line 69: feel like it makes sense to use a tx send helper or write one if it doesn't exist already. I'm sure we have a number of places where this would be relevant

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -2,19 +2,33 @@
...
+};
+use renegade_common::types::chain::Chain;
+use tracing::{info, warn};
 
 use crate::helpers::IERC20;
 
 use super::{error::ExecutionClientError, ExecutionClient};
 
+/// The factor by which the swap size will be divided when retrying
+const SWAP_SIZE_DIVISOR: U256 = U256::from_limbs([2, 0, 0, 0]);
```
joeykraut, line 24: maybe `SWAP_DECAY_FACTOR` ? The current name is a little unclear to me

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -65,22 +101,54 @@ impl ExecutionClient {
...
         if !receipt.status() {
-            let error_msg = format!("tx ({:#x}) failed", receipt.transaction_hash);
-            return Err(ExecutionClientError::arbitrum(error_msg));
+            warn!("tx ({:#x}) failed, retrying w/ reduced-size quote", receipt.transaction_hash);
+
+            let reduced_size_params =
+                LiFiQuoteParams { from_amount: params.from_amount / SWAP_SIZE_DIVISOR, ..params };
+
+            return Box::pin(self.swap_immediate_decaying(chain, reduced_size_params, wallet))
+                .await;
```
joeykraut, line 144: any reason not to just make this iterative? recursive futures are always a pain. Not super opinionated here though so feel free to ignore

joeykraut (approved):
very nice

funds-manager/funds-manager-server/src/execution_client/swap.rs
akirillo: ah, i was straying away from that name because i felt that "factor" implies multiplication, but i suppose we can rely on the doc comment for clarity

joeykraut: eh IMO it doesn't imply multiplication so much as the effect it has on the value: decaying at a constant factor. so seems like a fine name to me

PR #252, opened 2025-06-06
funds-manager: log LiFi tool used in quote

akirillo:
This PR logs the "tool" used by LiFi in a quote. We leave tagging metrics with this as a TODO.


funds-manager/funds-manager-server/src/execution_client/quotes.rs
```diff
@@ -19,6 +20,9 @@ impl ExecutionClient {
         let url = format!("{}?{}", QUOTE_ENDPOINT, query_string);
 
         let resp: LiFiQuote = self.send_get_request(&url).await?;
+
+        info!("Got LiFi quote from: {}", resp.tool);
```
sehyunc, line 24: is it possible to also log the tool alongside the swap details, maybe with the "Swap executed at ..." as metadata?

sehyunc (approved):

PR #257, opened 2025-06-10
auth-server: api-handlers: order-book: Add "all pairs" order book helper

joeykraut:
### Purpose
This PR adds a proxy endpoint for the `/v0/order_book/depth` endpoint on the relayer which returns book depth for all supported pairs.

### Testing
- [x] Tested locally
- [ ] Test in testnet


sehyunc (approved):
sweet, a proxy for supported tokens and their prices would be nice as well (in a separate PR)

PR #259, opened 2025-06-11
funds-manager: execution-client: swap: Add better request tracing

joeykraut:
### Purpose
This PR adds fine grained request tracing to swap requests in the funds manager. This will help us understand and debug the recursive trade execution better.

I also added a small optimization to over-approve ERC20 tokens by a constant amplification factor. This removes the need for sending an approval on _every_ swap.

### Testing
- [ ] Will test in mainnet where this path is active.

### DIFF
--- a/funds-manager/funds-manager-server/src/execution_client/mod.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/mod.rs
@@ -93,7 +93,7 @@ impl ExecutionClient {
     }
 
     /// Send a transaction, awaiting its receipt
-    #[instrument(skip(self, tx, client))]
+    #[instrument(skip_all)]
     async fn send_tx(
         &self,
         tx: TransactionRequest,
--- a/funds-manager/funds-manager-server/src/execution_client/mod.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/mod.rs
@@ -12,9 +12,10 @@ use alloy::{
     signers::local::PrivateKeySigner,
 };
 use http::StatusCode;
+use renegade_util::telemetry::helpers::backfill_trace_field;
 use reqwest::{Client, Url};
 use serde::Deserialize;
-use tracing::error;
+use tracing::{error, instrument};
 
 use crate::helpers::build_provider;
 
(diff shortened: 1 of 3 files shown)


funds-manager/funds-manager-server/src/execution_client/mod.rs
```diff
@@ -92,6 +93,7 @@ impl ExecutionClient {
     }
 
     /// Send a transaction, awaiting its receipt
+    #[instrument(skip(self, tx, client))]
```
akirillo, line 96: nit/fyi: can just do `instrument(skip_all)`, a little simpler

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -24,6 +24,10 @@ use super::{error::ExecutionClientError, ExecutionClient};
 const SWAP_DECAY_FACTOR: U256 = U256::from_limbs([2, 0, 0, 0]);
 /// The minimum amount of USDC that will be attempted to be swapped recursively
 const MIN_SWAP_QUOTE_AMOUNT: f64 = 10.0; // 10 USDC
+/// The amount to increase an approval by for a swap
+///
+/// We "over-approve" so that we don't need to re-approve on every swap
+const APPROVAL_AMPLIFIER: u64 = 4;
```
akirillo, line 30: just so we're on the same page: we only send an approval TX for the top-level amount requested, not for every "decaying" swap. does that mean this over-approval is intended to minimize the total # of approvals across multiple calls to the `/swap-immediate` endpoint?

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -123,7 +128,8 @@ impl ExecutionClient {
         }
 
         // Otherwise, approve the allowance
-        let tx = erc20.approve(spender, amount);
+        let approval_amount = amount * U256::from(APPROVAL_AMPLIFIER);
```
akirillo, line 131: nit: we can make `APPROVAL_AMPLIFIER` a constant `U256` by using `U256::from_limbs`, which is a `const fn`

akirillo (approved):

funds-manager/funds-manager-server/src/execution_client/swap.rs
joeykraut: yes correct

PR #261, opened 2025-06-11
funds-manager: execution_client: swap: Track cumulative gas cost

joeykraut:
### Purpose
This PR tracks the cumulative gas cost in the swap execution metrics. I only log for now, and don't record the metrics because I want to test this value in mainnet before we start recording real metrics to the NAV dashboard.

### Testing
- [ ] Will test in mainnet

### DIFF
--- a/funds-manager/funds-manager-server/src/metrics/cost.rs
+++ b/funds-manager/funds-manager-server/src/metrics/cost.rs
@@ -4,7 +4,8 @@ use alloy::rpc::types::TransactionReceipt;
 use alloy_primitives::{Address, Log, TxHash, U256};
 use alloy_sol_types::SolEvent;
 use funds_manager_api::{quoters::AugmentedExecutionQuote, u256_try_into_u128};
-use renegade_common::types::chain::Chain;
+use renegade_common::types::{chain::Chain, token::Token};
+use renegade_darkpool_client::conversion::u256_to_amount;
 use serde::Serialize;
 use tracing::{info, warn};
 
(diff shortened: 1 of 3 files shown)


funds-manager/funds-manager-server/src/metrics/cost.rs
```diff
@@ -228,6 +235,21 @@ impl MetricsRecorder {
         Ok(price)
     }
 
+    /// Get the cost of the given amount of WEI in USD
+    async fn get_wei_cost_usdc(&self, wei: U256) -> Result<f64, FundsManagerError> {
+        // We use the weth price here as a stand in for the eth price seeing as weth
+        // does not trade at a discount to eth. As well, we fetch the price using the
+        // arbitrum one chain for simplicity -- the price will be the same across chains
+        let weth = Token::from_ticker("WETH");
```
akirillo, line 243: let's be sure to use `Token::from_ticker_on_chain("WETH", Chain::ArbitrumOne)` here - this ensures that the `Token` constructed by the `PriceReporterClient` using `weth.get_alloy_address()` matches the one we constructed here.

not strictly necessary currently - it wouldn't result in any errors if we construct the Base WETH token and try getting its price while passing in `Chain::ArbitrumOne` - but helps protect against regressions if we use the `chain` parameter for anything else in `get_price`

funds-manager/funds-manager-server/src/metrics/cost.rs
```diff
@@ -56,6 +57,8 @@ pub struct SwapExecutionData {
     pub relative_spread: f64,
     /// The execution cost in USDC
     pub execution_cost_usdc: f64,
+    /// The gas cost of execution in WEI
```
akirillo, line 60: nit: 
```suggestion
    /// The gas cost of execution in USD
```

akirillo (approved):

PR #263, opened 2025-06-13
funds-manager: handlers: Return execution cost in `/swap-immediate`

joeykraut:
### Purpose
This PR returns execution cost in the API response from the funds manager on the `/swap-immediate` route. This cost represents the USD denominated cost of executing on LiFi versus Binance for the swap.

This will be used in the bot server to enable per-asset execution cost rate limits.

I moved the metrics recorder out of a thread to enable using the swap costs in the hot path. This operation is not latency expensive, and the endpoint in generally does not have stringent latency constraints on it. 

### Testing
- [ ] Have to test in mainnet, swaps only happen there

### DIFF
--- a/funds-manager/funds-manager-server/src/handlers.rs
+++ b/funds-manager/funds-manager-server/src/handlers.rs
@@ -250,17 +250,21 @@ pub(crate) async fn swap_immediate_handler(
     let (augmented_quote, receipt, swap_gas_cost) =
         execution_client.swap_immediate_decaying(chain, params, wallet).await?;
 
-    let resp = SwapImmediateResponse {
+    // Compute swap costs and respond
+    let execution_cost =
+        match metrics_recorder.record_swap_cost(&receipt, &augmented_quote, swap_gas_cost).await {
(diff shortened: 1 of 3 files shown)


funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -250,17 +250,16 @@ pub(crate) async fn swap_immediate_handler(
...
-    };
-
-    // Record swap cost metrics
-    tokio::spawn(async move {
-        metrics_recorder.record_swap_cost(&receipt, &augmented_quote, swap_gas_cost).await;
-    });
-
-    Ok(warp::reply::json(&resp))
+        execution_cost,
+    }))
```
akirillo, line 262: IMO, if we fail to record swap costs, we shouldn't fail to send a successful response here. The bot server should still be able to observe a successful swap to proceed w/ its rebalancing logic in the even that execution costs aren't successfully added to the response

akirillo (approved):
lgtm w/ one small comment

PR #266, opened 2025-06-14
auth-server: record price staleness metric

sehyunc:
### Purpose
This PR adds metrics recording for the price staleness of a bundle. Price staleness is calculated as the time between time of price sample and the time of settlement. For direct match bundles, which do not have a TimestampedPrice, we approximate by using the time the post-request subroutines are run. 

Metric added is `auth_server.external_match_price_staleness` in milliseconds.

### Testing
- [x] Test locally that price staleness is correctly calculated for basic match
- [ ] Test malleable match + direct match

### DIFF
--- a/auth/auth-server/src/chain_events/tasks.rs
+++ b/auth/auth-server/src/chain_events/tasks.rs
@@ -8,7 +8,7 @@ use renegade_circuit_types::order::OrderSide;
 use renegade_common::types::token::Token;
 use renegade_darkpool_client::DarkpoolClient;
 
-use crate::telemetry::labels::EXTERNAL_MATCH_PRICE_STALENESS;
+use crate::telemetry::labels::EXTERNAL_MATCH_SETTLEMENT_DELAY;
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/chain_events/tasks.rs
```diff
@@ -171,6 +175,40 @@ impl OnChainEventListenerExecutor {
         Ok(())
     }
 
+    /// Record the time between the canonical exchange midpoint sample time and
+    /// the time of settlement
+    pub async fn record_price_staleness(
```
joeykraut, line 180: nit: I'd call this `record_settlement_delay`, that seems a more descriptive name.

auth/auth-server/src/chain_events/tasks.rs
```diff
@@ -171,6 +175,40 @@ impl OnChainEventListenerExecutor {
...
+        &self,
+        ctx: &BundleContext,
+        tx: TxHash,
```
joeykraut, line 183: nit: I would reverse these two args, `tx` is the most important. The other two are more contextual

joeykraut (approved):

PR #271, opened 2025-06-25
auth-server: record price -> assemble -> settle delay metrics

sehyunc:
### Purpose
This PR adds metrics recording for time of price sample -> time of assembly -> time of settlement.

### Testing
- [x] Tested locally
- [ ] Test in testnet

### DIFF
--- a/auth/auth-server/src/chain_events/tasks.rs
+++ b/auth/auth-server/src/chain_events/tasks.rs
@@ -6,7 +6,6 @@ use bigdecimal::{BigDecimal, FromPrimitive, ToPrimitive};
 use renegade_api::http::external_match::ApiExternalMatchResult;
 use renegade_circuit_types::order::OrderSide;
 use renegade_common::types::token::Token;
-use renegade_darkpool_client::DarkpoolClient;
 
 use crate::chain_events::utils::GPv2Settlement;
 use crate::{bundle_store::BundleContext, chain_events::listener::OnChainEventListenerExecutor};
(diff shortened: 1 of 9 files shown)


auth/auth-server/src/chain_events/tasks.rs
```diff
@@ -240,4 +250,38 @@ impl OnChainEventListenerExecutor {
...
+        let settlement_time = block.header.timestamp * 1000; // Convert to milliseconds
+
+        Ok(settlement_time)
+    }
+
+    /// Calculate and record the time between two timestamps
+    fn record_time_diff(
+        &self,
+        t_1: u64,
+        t_2: u64,
```
joeykraut, line 280: style nit: in keeping consistent with the rest of our codebases, we don't usually do `t_1` and instead prefer `t1`

auth/auth-server/src/server/api_handlers/external_match/mod.rs
```diff
@@ -20,6 +20,11 @@ pub(crate) use assemble_malleable_quote::SponsoredAssembleMalleableQuoteResponse
 
 use super::{get_sdk_version, log_unsuccessful_relayer_request};
 
+/// Returns the number of milliseconds since the Unix epoch
+pub(crate) fn get_timestamp_ms() -> u64 {
+    (chrono::Utc::now().timestamp() as u64) * 1000
+}
+
```
joeykraut, line 27: you have the `renegade_util::get_current_timestamp_millis` in scope, probably best to just use that helper rather than redefine it here.

auth/auth-server/src/server/api_handlers/settlement.rs
```diff
@@ -61,6 +63,7 @@ impl Server {
     /// Write the bundle context for a malleable match to the store
     pub async fn write_malleable_bundle_context(
         &self,
+        assembled_timestamp: Option<u64>,
```
joeykraut, line 66: this is only ever passed as `Some(..)`, I'd propose we just make it non-optional.

joeykraut (approved):

auth/auth-server/src/server/api_handlers/settlement.rs
sehyunc: In the case of direct matches, this is passed as `None` since they are not "assembled"

PR #272, opened 2025-06-26
auth-server: setup: Create execution cost rate limiter

joeykraut:
### Purpose
This PR adds a Redis-based execution cost rate limiter to the auth server. This rate limiter will read from the redis cluster written to by the bot server when it reports its execution costs.

Notably we don't fail the API execution if the rate limit check fails, we emit a warning for now. I will change this later if we see fit, but I'm planning to test this with live rate limits before changing routing behavior.

Routing decisions will come in a follow-up.

### Testing
- [x] Tested locally

### DIFF
--- a/auth/auth-server/src/server/rate_limiter/execution_cost_rate_limiter.rs
+++ b/auth/auth-server/src/server/rate_limiter/execution_cost_rate_limiter.rs
@@ -8,6 +8,7 @@ use crate::{error::AuthServerError, server::db::create_redis_client};
 
 /// A rate limiter that measures the bot server's per-asset execution costs and
 /// rate limits external match flow that crosses with quoter orders
+#[derive(Clone)]
 pub struct ExecutionCostRateLimiter {
     /// The Redis connection manager
     redis: RedisConnection,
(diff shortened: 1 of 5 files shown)


auth/auth-server/src/server/rate_limiter/execution_cost_rate_limiter.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// Constructor
+    pub async fn new(redis_url: &str) -> Result<Self, AuthServerError> {
+        let conn = create_redis_client(redis_url).await?;
+        Ok(Self { redis: conn })
+    }
+
+    /// Check if the rate limit has been exceeded for the given ticker
+    pub async fn rate_limit_exceeded(&self, ticker: &str) -> Result<bool, AuthServerError> {
+        let key = Self::get_rate_limit_exceeded_key(ticker);
+        let result: Option<bool> = self.redis().get(key).await?;
```
akirillo, line 27: iirc we write the value for this as either a `0` or a `1`, just want to confirm that the Rust redis client will indeed parse this as a `bool`?

akirillo (approved):

auth/auth-server/src/server/rate_limiter/execution_cost_rate_limiter.rs
joeykraut: yes i've confirmed in my local testing

PR #273, opened 2025-06-26
auth-server: api-handlers: Check execution rate limits before forwarding

joeykraut:
### Purpose
This PR adds logic in the pre-request handlers for each of the external match endpoints to check the execution rate limit for the pair they're matching. For now, if the rate limit is exceeded, the auth server will just log a warning. In a follow-up I'll add routing logic that places orders in the global pool after rate limits are exceeded.

### Testing
- [ ] Testing locally
- [ ] Testing in testnet

### DIFF
--- a/auth/auth-server/src/server/api_handlers/external_match/quote.rs
+++ b/auth/auth-server/src/server/api_handlers/external_match/quote.rs
@@ -5,7 +5,7 @@ use bytes::Bytes;
 use http::{Response, StatusCode};
 use renegade_api::http::external_match::{ExternalQuoteRequest, ExternalQuoteResponse};
 use renegade_circuit_types::fixed_point::FixedPoint;
-use renegade_common::types::{token::Token, TimestampedPrice};
+use renegade_common::types::{price::TimestampedPrice, token::Token};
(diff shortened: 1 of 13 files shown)


auth/auth-server/src/server/api_handlers/external_match/quote.rs
```diff
@@ -5,7 +5,7 @@ use bytes::Bytes;
 use http::{Response, StatusCode};
 use renegade_api::http::external_match::{ExternalQuoteRequest, ExternalQuoteResponse};
 use renegade_circuit_types::fixed_point::FixedPoint;
-use renegade_common::types::{token::Token, TimestampedPrice};
+use renegade_common::types::{price::TimestampedPrice, token::Token};
```
akirillo, line 8: getting an import warning from clippy here

akirillo (approved):

auth/auth-server/src/server/api_handlers/external_match/quote.rs
joeykraut: forgot to push the Cargo.lock update, thanks

PR #275, opened 2025-07-01
auth-server: Handle bad request errors properly

joeykraut:
### Purpose
This PR adds a rejection handler branch for the `AuthServerError` type and adds a `BadRequest` variant to this error. We use this error type to return HTTP 400 when an address is not in the token mapping. 

Previously the auth server would return an opaque "500: Internal Server Error" because the `AuthServerError` type was not handled.

### Testing
- [x] Tested locally with an invalid token address.

### DIFF
--- a/auth/auth-server/src/main.rs
+++ b/auth/auth-server/src/main.rs
@@ -34,9 +34,14 @@ use std::sync::Arc;
 use thiserror::Error;
 use tracing::{error, info, info_span};
 use uuid::Uuid;
-use warp::{Filter, Rejection, Reply};
+use warp::{
+    reply::{Json, WithStatus},
+    Filter, Rejection,
+};
 
 use server::Server;
+
+use crate::error::AuthServerError;
 
 /// The default internal server error message
 const DEFAULT_INTERNAL_SERVER_ERROR_MESSAGE: &str = "Internal Server Error";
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/main.rs
```diff
@@ -363,33 +368,43 @@ fn with_tracing() -> warp::trace::Trace<impl Fn(warp::trace::Info) -> tracing::S
...
-                error!("Internal server error: {e}");
-                (StatusCode::INTERNAL_SERVER_ERROR, DEFAULT_INTERNAL_SERVER_ERROR_MESSAGE)
-            },
-            ApiError::BadRequest(msg) => (StatusCode::BAD_REQUEST, msg.as_str()),
-            ApiError::TooManyRequests => (StatusCode::TOO_MANY_REQUESTS, "Rate limit exceeded"),
-            ApiError::Unauthorized => (StatusCode::UNAUTHORIZED, "Unauthorized"),
-        };
-
-        Ok(json_error(message, code))
+async fn handle_rejection(err: Rejection) -> Result<WithStatus<Json>, Rejection> {
```
sehyunc, line 371: very cool to handle API layer errors by matching and a `json_error` constructor, it's giving me ideas of how to do this in js.

PR #278, opened 2025-07-01
auth-server: api-handlers: Add `GET /api-keys` method

joeykraut:
### Purpose
This PR adds a `GET /api-keys` method which returns all API keys registered to the auth server. This will be used in the admin panel to view API keys and their whitelist status.

### Testing
- [x] Tested locally

### DIFF
--- a/auth/auth-server-api/src/key_management.rs
+++ b/auth/auth-server-api/src/key_management.rs
@@ -1,0 +1,30 @@
+//! Key management API endpoints
+
+use std::time::SystemTime;
+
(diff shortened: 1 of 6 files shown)


sehyunc (approved):
we should invalidate the cache in `whitelist_api_key_query` and `remove_whitelist_entry_query`

sehyunc:
> we should invalidate the cache in `whitelist_api_key_query` and `remove_whitelist_entry_query`

😪 nvm I see #279

PR #279, opened 2025-07-01
auth-server: api-handlers: Do not re-route whitelisted api keys

joeykraut:
### Purpose
This PR checks the whitelist status for an API key before routing the order to the global matching pool. If an API key has the whitelist flag set, it will never be routed to the global matching pool. Orders without the whitelist flag set will still be routed according to the rate limits reported by the bot server.

### Testing
- [x] Tested whitelisted/non-whitelisted with rate-limited/non-rate-limited locally on my personal key.

### DIFF
--- a/auth/auth-server/src/server/api_handlers/key_management.rs
+++ b/auth/auth-server/src/server/api_handlers/key_management.rs
@@ -97,6 +97,8 @@ impl Server {
         // Check management auth on the request
         self.authorize_management_request(&path, &headers, &body)?;
         self.whitelist_api_key_query(key_id).await?;
+        self.clear_cached_key(key_id).await;
+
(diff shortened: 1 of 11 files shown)


auth/auth-server/src/server/api_handlers/key_management.rs
```diff
@@ -115,6 +117,8 @@ impl Server {
         // Check management auth on the request
         self.authorize_management_request(&path, &headers, &body)?;
         self.remove_whitelist_entry_query(key_id).await?;
+        self.clear_cached_key(key_id).await;
```
sehyunc, line 120: nice

sehyunc (approved):

PR #282, opened 2025-07-04
price_reporter_client: optional exit-on-stale

akirillo:
This PR adds an `exit_on_stale` option to the `PriceReporterClient` which, when set, has the client maintain a 1-minute timer that exits the current process. This timer gets re-created every time a price is updated, so as long as we're receiving prices within each minute, the self-destruct won't be invoked.

### Testing
- [x] Run local price reporter client w/ sleeps in between price updates, asserting that the process exits when the staleness timeout is hit
- [x] Run unmodified local price reporter client, asserting that prices are updated and the process runs w/o exiting

### DIFF
--- a/auth/auth-server/src/server/setup.rs
+++ b/auth/auth-server/src/server/setup.rs
@@ -72,8 +72,10 @@ impl Server {
         )
         .await?;
 
-        let price_reporter_client =
-            Arc::new(PriceReporterClient::new(args.price_reporter_url.clone())?);
+        let price_reporter_client = Arc::new(PriceReporterClient::new(
+            args.price_reporter_url.clone(),
+            false, // exit_on_stale
+        )?);
 
         // Setup quote metrics
         let quote_metrics = maybe_setup_quote_metrics(
--- a/price-reporter-client/src/price_stream.rs
+++ b/price-reporter-client/src/price_stream.rs
@@ -16,7 +16,7 @@ use futures_util::{
 };
 use renegade_api::websocket::WebsocketMessage;
 use serde::Deserialize;
-use tokio::{net::TcpStream, sync::RwLock};
+use tokio::{net::TcpStream, sync::RwLock, task::JoinHandle};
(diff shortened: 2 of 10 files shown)


auth/auth-server/src/server/setup.rs
```diff
@@ -72,8 +72,10 @@ impl Server {
         )
         .await?;
 
-        let price_reporter_client =
-            Arc::new(PriceReporterClient::new(args.price_reporter_url.clone())?);
+        let price_reporter_client = Arc::new(PriceReporterClient::new(
+            args.price_reporter_url.clone(),
+            false, // exit_on_stale
```
joeykraut, line 77: now that i'm looking at this, i think false is the right call for the auth server

price-reporter-client/src/price_stream.rs
```diff
@@ -29,6 +29,9 @@ use super::{construct_price_topic, error::PriceReporterClientError, get_base_min
 /// The number of milliseconds to wait in between retrying connections
 pub const CONN_RETRY_DELAY_MS: u64 = 2_000; // 2 seconds
 
+/// The timeout in milliseconds after which the price stream is considered stale
```
joeykraut, line 32: can you define what staleness is here? specifically is it "no price updates" or "no price updates per asset", etc

price-reporter-client/src/price_stream.rs
```diff
@@ -75,12 +87,44 @@ impl MultiPriceStreamState {
             .entry(mint)
             .or_insert(AtomicF64::new(price))
             .store(price, Ordering::Relaxed);
+
+        if self.exit_on_stale {
+            self.restart_staleness_timer().await;
+        }
```
joeykraut, line 93: nit: I'd encapsulate this flag entirely in the `restart_staleness_timer` method; i.e. that method just short-circuits if `self.exit_on_stale == false`. Slightly cleaner

joeykraut (approved):

PR #285, opened 2025-07-11
price-reporter: exchanges: Internalize exchange connection logic

joeykraut:
### Purpose
This PR internalizes the exchange connection logic into the price reporter. The dependency on the renegade price reporter will be removed in a follow-up.

This is merged into a feature branch for now.

### Todo
- Http server needs upgrade in the price reporter, will test after that is fixed


sehyunc (approved):
lgtm

PR #287, opened 2025-07-11
auth-server: api-handlers: Stringify response if `Accept: number=string`

joeykraut:
### Purpose
This PR checks the `Accept` header for each request type and string-serialized the response (following #286 ) if `number=string` is set in the header. 

### Testing
- [x] Tested with a local client setting `Accept: application/json;number=string"`.


sehyunc (approved):
so awesome that this is all it takes for stringified numbers

i may rely on this in our internal repos to remove a dependency

PR #288, opened 2025-07-14
auth-server: api-handlers: Allow stringified numbers in request

joeykraut:
### Purpose
This PR adds a `json_deserialize` method which can optionally convert all stringified numbers into number types ahead of deserialization. This allows clients to send numbers as string values.

### Testing
- [x] Tested on a locally running auth server
- [x] Testing in testnet

### DIFF
--- a/auth/auth-server/Cargo.toml
+++ b/auth/auth-server/Cargo.toml
@@ -70,7 +70,7 @@ num-bigint = "0.4"
 atomic_float = "1"
 rustls = "0.23"
 serde = { version = "1.0.218", features = ["derive"] }
-serde_json = "1.0.139"
+serde_json = { version = "1.0.139", features = ["arbitrary_precision"] }
 serde_urlencoded = "0.7"
 thiserror = "1.0"
 tracing = "0.1"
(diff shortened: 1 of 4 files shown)


joeykraut:
> got it, so requests from client to the auth server were failling to parse since numbers were expected instead of strings. this all lgtm

Basically, although this was intentional. The original plan was to only have `Accept: number=string` signal that the _response_ should be stringified, but it's more ergonomic if both request and response are stringified.

sehyunc:
> > got it, so requests from client to the auth server were failling to parse since numbers were expected instead of strings. this all lgtm
> 
> Basically, although this was intentional. The original plan was to only have `Accept: number=string` signal that the _response_ should be stringified, but it's more ergonomic if both request and response are stringified.

Right because the client essentially forwards the unchanged quote

joeykraut:
> > > got it, so requests from client to the auth server were failling to parse since numbers were expected instead of strings. this all lgtm
> > 
> > 
> > Basically, although this was intentional. The original plan was to only have `Accept: number=string` signal that the _response_ should be stringified, but it's more ergonomic if both request and response are stringified.
> 
> Right because the client essentially forwards the unchanged quote

yep, exactly

PR #294, opened 2025-07-22
funds-manager: handlers: add swap-into-target-token handler

akirillo:
This PR adds a handler & exposes a route for the new `try_swap_into_target_token` method.

Testing will have to be done on mainnet w/ small amounts, and is blocked by client-side implementation of this endpoint in Typescript. Any resultant bugfixes will come in a follow-up PR.

### DIFF
--- a/funds-manager/funds-manager-api/src/types/quoters.rs
+++ b/funds-manager/funds-manager-api/src/types/quoters.rs
@@ -35,6 +35,12 @@ pub const WITHDRAW_TO_HYPERLIQUID_ROUTE: &str = "withdraw-to-hyperliquid";
 /// - order: Order preference for routing (e.g. 'CHEAPEST')
 /// - slippage: Slippage tolerance as a decimal (e.g. 0.0001 for 0.01%)
 pub const SWAP_IMMEDIATE_ROUTE: &str = "swap-immediate";
+/// The route to execute swaps to cover a target amount of a given token.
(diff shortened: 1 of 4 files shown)


funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -297,6 +303,17 @@ pub struct SwapImmediateResponse {
...
 
+/// The request body for executing a swap to cover a target amount of a given
+/// token
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SwapIntoTargetTokenRequest {
+    /// The target amount of the token to cover, in decimal format (i.e., whole
+    /// units)
+    pub target_amount: f64,
+    /// The quote parameters for the swap
+    pub quote_params: LiFiQuoteParams,
```
joeykraut, line 314: What's the plan here? This value requires both a to and a from token whereas in reality the funds manager will decide the from token. Do we expect the from_token to be set?

joeykraut (approved):

funds-manager/funds-manager-api/src/types/quoters.rs
akirillo: on second thought, it's simpler to expect `from_token` to be set, e.g. to the zero address. this will get overwritten anyway and leads to less type wrangling

PR #295, opened 2025-07-22
price-reporter: exchanges: coinbase: Do not recompute book each message

joeykraut:
### Purpose
This PR updates the Coinbase connection handler to persist order book data across messages. Previously a buggy implementation meant that we recomputed the book anew from each message. However, each message only contains book updates, so we were recomputing the book based on only new placements/cancellations. 

This is likely the cause of the price flickering as new price levels entirely determine the new book.

I also refactored the order book data implementation to use non-blocking crossbeam atomic primitives under the hood, so that it is more efficient with potentially concurrent updates.

### Testing
- [x] Tested locally
- [x] Testing in testnet

### DIFF
--- a/price-reporter/src/exchanges/coinbase.rs
+++ b/price-reporter/src/exchanges/coinbase.rs
@@ -1,14 +1,16 @@
 //! Defines handler logic for a Coinbase websocket connection
 
 use std::{
-    collections::HashMap,
     pin::Pin,
+    sync::Arc,
     task::{Context, Poll},
 };
 
 use async_trait::async_trait;
+use crossbeam_skiplist::SkipSet;
 use futures_util::{Sink, SinkExt, Stream, StreamExt};
 use jsonwebtoken::{encode, Algorithm, EncodingKey as JwtEncodingKey, Header as JwtHeader};
+use ordered_float::NotNan;
 use renegade_common::types::{exchange::Exchange, price::Price, token::Token};
 use renegade_util::{err_str, get_current_time_seconds};
 use reqwest::{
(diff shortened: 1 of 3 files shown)


price-reporter/src/exchanges/coinbase.rs
```diff
@@ -323,3 +296,83 @@ impl ExchangeConnection for CoinbaseConnection {
...
 }
+
+// ------------------
+// | Orderbook Data |
+// ------------------
+
+/// A non-nan f64
+type NonNanF64 = NotNan<f64>;
+/// A shared skip set of price levels
+pub type OrderBookLevels = Arc<SkipSet<NonNanF64>>;
```
sehyunc, line 307: IIUC: `SkipSet` for fast `.front()` and `.back()` and because we only care about existence of a price level rather than the actual depth?

price-reporter/src/exchanges/coinbase.rs
```diff
@@ -323,3 +296,83 @@ impl ExchangeConnection for CoinbaseConnection {
...
+
+    /// Get the best offer price from the current order book
+    pub fn best_offer(&self) -> Option<f64> {
+        self.offers.front().map(|e| e.value().into_inner())
+    }
+
+    /// Get the midpoint price from the current order book
+    pub fn midpoint(&self) -> Option<f64> {
+        let best_bid = self.best_bid()?;
+        let best_offer = self.best_offer()?;
```
sehyunc, line 343: Can we constantize these, maybe like `EMPTY_BID` or something

sehyunc (approved):
very cool and clean

price-reporter/src/exchanges/coinbase.rs
joeykraut: Removed them

price-reporter/src/exchanges/coinbase.rs
joeykraut: That's correct

PR #298, opened 2025-07-23
auth-server: api-handlers: Implement fee override managment interface

joeykraut:
### Purpose
This PR sets up API handlers for:
- Getting the current external match fees on a per-user, per-asset basis.
- Setting a default fee for an asset
- Setting a per-user fee override

### Testing
- [x] Tested all endpoints locally against the testnet DB

### DIFF
--- a/auth/auth-server/src/server/db/queries.rs
+++ b/auth/auth-server/src/server/db/queries.rs
@@ -7,8 +7,10 @@ use uuid::Uuid;
 use crate::{error::AuthServerError, server::Server};
 
 use super::{
-    models::{ApiKey, NewApiKey},
-    schema::api_keys,
+    models::{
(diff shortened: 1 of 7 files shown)


auth/auth-server/src/server/db/queries.rs
```diff
@@ -120,4 +122,98 @@ impl Server {
...
+            .await
+            .map_err(AuthServerError::db)
+    }
+
+    /// Get the cartesian product of active users and assets with fee
+    /// inheritance
+    ///
+    /// This joins active API keys with asset default fees and left joins with
+    /// user overrides. It also includes assets that only have user overrides
+    /// but no default fees.
```
sehyunc, line 186: Will there be assets with neither a default fee nor a user override fee?

sehyunc (approved):

PR #302, opened 2025-07-24
auth-server: caching: Add per-user fee cache

joeykraut:
### Purpose
This PR adds a cache for per-user fees. This will allow us to quickly determine the fee the relayer should charge a user on a per-asset basis. We will populate this cache in a follow up with a new method that requests the user's fee falling back to the default for an asset (or zero if neither is set).

I also changed the underlying cache structure to a `DashMap` which generally has much better performance under load.

### Testing
- Deferred until caching implementation is in use in the next PR


auth/auth-server/src/server/db/queries.rs
```diff
@@ -167,7 +176,46 @@ impl Server {
         Ok(())
     }
 
-    // --- Asset Default Fees --- //
+    /// Get the cartesian product of active users and assets with fee
```
joeykraut, line 179: moved from below

PR #303, opened 2025-07-24
auth-sever: api-handlers: Set relayer fee in request processing

joeykraut:
### Purpose
This PR sets the `relayer_fee` field on each request during preprocessing according to the fee override stored in the DB.

### Testing
- [x] Testing locally
- [ ] Testing in testnet


sehyunc (approved):
very clean lgtm

PR #304, opened 2025-07-25
solver: add executor client

sehyunc:
### Purpose
This PR adds an ExecutorClient to interact with the DarkpoolExecutor contract.

### Testing
- Tested is deferred until the contract is deployed


renegade-solver/src/uniswapx/executor_client/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// The address of the executor proxy contract.
+    ///
+    /// This is the main entrypoint to interaction with the executor.
+    pub contract_address: String,
+    /// HTTP-addressable RPC endpoint for the client to connect to
+    pub rpc_url: String,
+    /// The private key of the account to use for signing transactions
+    pub private_key: PrivateKeySigner,
+    /// The interval at which to poll for event filters and pending transactions
+    pub block_polling_interval: Option<Duration>,
```
joeykraut, line 39: does this need to be configured? are we realistically going to change this between deployments? I don't personally think so, and if not it's much more ergonomic to just constantize a reasonable value. 

FWIW I think it's a mistake that this is configurable in the `darkpool-client`. Large configs are a huge problem

renegade-solver/src/uniswapx/executor_client/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        let latest_block = self
+            .provider()
+            .get_block(BlockId::latest())
+            .await
+            .map_err(ExecutorError::rpc)?
+            .ok_or(ExecutorError::rpc("No latest block found"))?;
+
+        let latest_basefee =
+            latest_block.header.base_fee_per_gas.ok_or(ExecutorError::rpc("No basefee found"))?;
+        let gas_price = (latest_basefee * 2) as u128;
```
joeykraut, line 148: can we constatize this gain parameter up at the top of the file?

renegade-solver/src/uniswapx/executor_client/contract_interaction.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl ExecutorClient {
+    // -----------
+    // | GETTERS |
+    // -----------
+
+    /// Checks whether the given address is a whitelisted solver
+    pub async fn is_whitelisted_solver(&self, solver: Address) -> Result<bool, ExecutorError> {
+        Ok(self.contract.isWhitelistedSolver(solver).call().await?)
+    }
```
joeykraut, line 19: do we plan on using this method? Seems like we won't really need to be checking addresses ever. If not, we should remove it.

joeykraut (approved):

PR #308, opened 2025-07-29
solver: add amount scaling helpers

sehyunc:
### Purpose
This PR adds methods to scale the input or the output according to the `mpsPerPriorityFeeWei` value and the `priorityFee` bid. These methods are translated directly from their Solidity counterparts to ensure the bundle we fetch from the relayer is parameterized by the correct amounts to prevent reverts.

### Testing
- [x] Ran unit tests

### DIFF
--- a/renegade-solver/src/error.rs
+++ b/renegade-solver/src/error.rs
@@ -41,6 +41,9 @@ pub enum SolverError {
     /// Fixed point math error
     #[error("Fixed point math error: {0}")]
     FixedPoint(#[from] FixedPointMathError),
+    /// An order's input and outputs both scale with priority fee
+    #[error("Input and outputs both scale with priority fee")]
+    InputOutputScaling,
 }
 
 impl SolverError {
--- a/renegade-solver/src/uniswapx/renegade_api.rs
+++ b/renegade-solver/src/uniswapx/renegade_api.rs
@@ -16,12 +16,9 @@ impl UniswapXSolver {
     /// Renegade API. Validation for this should be done in the caller.
     pub(crate) async fn solve_renegade_leg(
         &self,
-        input_token: Address,
-        output_token: Address,
-        input_amount: u128,
+        order: ExternalOrder,
     ) -> SolverResult<Option<AtomicMatchApiBundle>> {
         let opt = Self::get_external_match_options();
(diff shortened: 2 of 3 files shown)


renegade-solver/src/uniswapx/abis/priority_order.rs
```diff
@@ -73,3 +84,70 @@ impl PriorityOrder {
...
+const MPS_U256: U256 = U256::from_limbs([MPS, 0, 0, 0]);
+
+impl PriorityInput {
+    /// Returns the input scaled by 1 - (priority_fee ×
+    /// priority_fee_scaling_rate)
+    ///
+    /// This implements scaling favoring the swapper where higher priority fees
+    /// result in lower required input amounts. The scaling is bounded by 0
+    /// and the original input amount.
+    pub fn scale(&self, priority_fee: U256) -> FixedPointResult<U256> {
```
joeykraut, line 101: nit: maybe call this parameter `priority_fee_wei` to indicate the expected units

renegade-solver/src/uniswapx/fixed_point/error.rs
```diff
@@ -0,0 +1,16 @@
+//! Error types for fixed point math
+use thiserror::Error;
+
+/// Type alias for Results using FixedPointMathError
+pub type FixedPointResult<T> = Result<T, FixedPointMathError>;
```
joeykraut, line 5: Is `T` ever a type other than `U256`? If not, we should just set `type FixedPointResult = Result<U256, FixedPointMathError>`

renegade-solver/src/uniswapx/fixed_point/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    if y.is_zero() {
+        return Ok(U256::ZERO);
+    }
+
+    // Check for overflow: x <= U256::MAX / y
+    let max_x = U256::MAX / y;
+    if x > max_x {
+        return Err(FixedPointMathError::Overflow);
+    }
```
joeykraut, line 44: maybe add a `fn validate_division` to reuse with the method above

joeykraut (approved):

PR #321, opened 2025-07-31
funds-manager: execution_client: use ExecutableQuote for Lifi execution

akirillo:
This PR completes porting over the current execution stack (Lifi-only) to using the new `ExecutionQuote` and `ExecutableQuote` types. This includes some refactoring, notably:
1. I removed the `AugmentedExecutionQuote` type, and moved its helper methods to `ExecutionQuote`. This involved some deduplication of adjacent methods on the old `ExecutionQuote` struct
2. I removed some fields from the `SwapCostData` struct, as they were not relied on in practice, and the fields from which they are computed no longer exist on the new `ExecutionQuote` struct, i.e. are not venue-agnostic.
3. I moved the calculation of `buy_amount_actual` from the `MetricsRecorder` to the `ExecutionClient`. This is largely for ergonomics, since execution venues may expose different ways to determine this amount. For example, Cowswap has a `/trades` endpoint that will report the actual buy amount on a trade instead of having to infer it from the TX logs

### Testing
- [x] Test swaps on local funds manager, assert that metrics and logs line up with what's expected

### DIFF
--- a/funds-manager/funds-manager-api/src/types/quoters.rs
+++ b/funds-manager/funds-manager-api/src/types/quoters.rs
@@ -1,15 +1,8 @@
 //! API types for quoter management
-use alloy_primitives::{hex, Address, Bytes, U256};
-use renegade_common::types::{
-    chain::Chain,
-    token::{Token, USDC_TICKER},
-};
+use alloy_primitives::U256;
 use serde::{Deserialize, Serialize};
 
-use crate::{
-    serialization::{f64_string_serialization, u256_string_serialization},
-    u256_try_into_u128,
-};
+use crate::serialization::{f64_string_serialization, u256_string_serialization};
 
 // --------------
 // | Api Routes |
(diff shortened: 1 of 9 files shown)


funds-manager/funds-manager-server/src/execution_client/venues/quote.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub sell_amount: U256,
+    /// The quoted amount of the token being bought, in atoms
+    pub buy_amount: U256,
+    /// The venue that provided the quote
+    pub venue: SupportedExecutionVenue,
+    /// The chain for which the quote was generated
+    pub chain: Chain,
+}
+
+impl ExecutionQuote {
```
akirillo, line 29: other than this `impl` block, the code in this file is just me moving the types over from `execution_client/venues/mod.rs`

funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -66,218 +59,23 @@ pub struct WithdrawFundsRequest {
 
 // --- Execution --- //
 
-/// The subset of the quote response forwarded to consumers of this client
-// TODO: Remove in favor of `funds_manager_server::execution_client::venues::ExecutionQuote`
-#[derive(Clone, Debug, Serialize, Deserialize)]
+/// A simplified representation of an execution quote, suitable for API
+/// responses
+#[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
```
joeykraut, line 65: very nice

joeykraut (approved):

PR #322, opened 2025-08-01
funds-manager: execution_client: ExecutionVenue trait + Lifi impl

akirillo:
This PR introduces the `ExecutionVenue` trait, through which we will get & execute quotes across all execution venues. We implement the trait for Lifi within a new `LifiClient` struct, which ingests much of the logic currently defined on the `ExecutionClient`.

In its current state, the `funds-manager` crate doesn't build. In the next PR, I'll implement proper initialization of the `LifiClient` and store it on the `ExecutionClient` - then, we'll be able to invoke it properly in the swap handlers.


funds-manager/funds-manager-server/src/execution_client/venues/lifi.rs
```diff
@@ -3,23 +3,145 @@
...
+const DEFAULT_DENIED_EXCHANGES: [&str; 1] = ["sushiswap"];
+
+// ---------
+// | Types |
+// ---------
+
+/// The subset of Lifi quote request query parameters that we support
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LifiQuoteParams {
```
akirillo, line 94: this was copied over from `funds-manager-api/src/types/quoters.rs`

funds-manager/funds-manager-server/src/helpers.rs
```diff
@@ -135,6 +140,41 @@ pub async fn get_erc20_balance(
     Ok(bal_f64)
 }
 
+/// Approve an erc20 allowance
+pub(crate) async fn approve_erc20_allowance(
```
akirillo, line 144: moved over from `funds-manager/funds-manager-server/src/execution_client/swap.rs`

PR #327, opened 2025-08-03
funds-manager-api: types: quoters: chain-aware is_sell method

akirillo:
This PR ensures that the `is_sell` method on the `ExecutionQuote` constructs a chain-aware USDC token. The current `Token::usdc()` call assumes the `default_chain`. This results in incorrect swap cost metrics, and also leads to quotes being considered invalid on acount of price deviation, because in sells on the non-default chain, the base token is always assumed to be USDC and have a price of 1.

I will be sure that these changes are reflected appropriately into the `andrew/multi-venue-quote-selection` branch

### Testing
- [x] Tested by executing swaps via local funds manager that were failing on the deployed version

### DIFF
--- a/funds-manager/funds-manager-api/src/types/quoters.rs
+++ b/funds-manager/funds-manager-api/src/types/quoters.rs
@@ -103,44 +103,6 @@ pub struct ExecutionQuote {
     pub gas_limit: U256,
 }
 
-impl ExecutionQuote {
-    /// Whether the quote is a sell order in Renegade terms
-    pub fn is_sell(&self) -> bool {
-        let usdc = Token::usdc();
-        self.buy_token_address == usdc.get_alloy_address()
-    }
-
-    /// Get the base token address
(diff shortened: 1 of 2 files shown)


funds-manager/funds-manager-api/src/types/quoters.rs
```diff
@@ -105,38 +105,38 @@ pub struct ExecutionQuote {
 
 impl ExecutionQuote {
     /// Whether the quote is a sell order in Renegade terms
-    pub fn is_sell(&self) -> bool {
-        let usdc = Token::usdc();
+    pub fn is_sell(&self, chain: Chain) -> bool {
+        let usdc = Token::from_ticker_on_chain(USDC_TICKER, chain);
         self.buy_token_address == usdc.get_alloy_address()
```
joeykraut, line 110: Probably better to move these to the augmented quote where chain is available immediately

joeykraut (approved):

PR #329, opened 2025-08-04
prover-service: Setup prover service

joeykraut:
### Purpose
This PR initializes the prover service; which will handle high volume proof workloads for the relayer.

This adds a dockerfile and a simple ping endpoint on the server. The implementation will somewhat straightforwardly follow from the existing proof manager worker in the relayer. This will come in a follow-up.

### Testing
- [x] Ran locally

### DIFF
--- a/prover-service/src/main.rs
+++ b/prover-service/src/main.rs
@@ -32,10 +32,8 @@ async fn main() {
     let cli = Cli::parse();
     cli.configure_telemetry().expect("failed to setup telemetry");
 
-    // --- Setup --- //
+    // Run the server
     let routes = setup_routes();
-
-    // --- Run --- //
     let listen_addr: SocketAddr = ([0, 0, 0, 0], cli.port).into();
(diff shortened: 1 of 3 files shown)


prover-service/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// --- Middleware --- //
+
+/// Custom tracing filter that creates spans for requests at info level
+/// with the auth_server::request target to work with our RUST_LOG configuration
+fn with_tracing() -> warp::trace::Trace<impl Fn(warp::trace::Info) -> tracing::Span + Clone> {
+    warp::trace(|info| {
+        let span = info_span!(
+            target: "auth_server::request",
```
akirillo, line 63:
```suggestion
            target: "prover_service::request",
```

akirillo: (and in the doc comment above)

akirillo (approved):

PR #331, opened 2025-08-04
prover-service: prover: Add endpoint handler structure

joeykraut:
### Purpose
This PR adds the endpoint handler structure for the prover service. This has one endpoint per proof that will be generated.

### Todo
- Add auth, we'll use HTTP basic auth for this service

### Testing
- [x] Tested locally

### DIFF
--- a/prover-service/server/src/prover.rs
+++ b/prover-service/server/src/prover.rs
@@ -1,0 +1,91 @@
+//! The prover service implementation
+
+// ---------------------
+// | Endpoint Handlers |
+// ---------------------
+
+use prover_service_api::{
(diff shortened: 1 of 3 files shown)


prover-service/server/src/prover.rs
akirillo, line 1: tiny nit: `prover.rs` is an abstract and misleading name for this file, can we call it `handlers.rs`?

akirillo (approved):

prover-service/server/src/prover.rs
joeykraut: Handler is more abstract than prover, prover gets at what the file _does_.

PR #336, opened 2025-08-05
prover-service: Add HTTP Basic Auth

joeykraut:
### Purpose
This PR adds HTTP basic authentication to all prover endpoints on the `prover-service`. 

### Testing
- [x] Tested server locally


prover-service/server/src/main.rs
```diff
@@ -32,16 +27,33 @@ use crate::{
 
 mod cli;
 mod error;
+mod middleware;
 mod prover;
 
+/// The runtime stack size to use for the server
+const RUNTIME_STACK_SIZE: usize = 50 * 1024 * 1024; // 50MB
```
akirillo, line 34: in my experience w/ the auth server & funds manager, i only need to bump the thread stack size when running the `dev` build, not `release`, so I've never committed this change. were you getting stack overflows on `release`?

joeykraut: no, see the comment i added, it's for debug mode

PR #344, opened 2025-08-07
funds-manager: helpers: send ERC20 approvals w/ a retry

akirillo:
This PR makes a couple fixes to mitigate nonce issues in the funds manager:

First, we remove usage of `ProviderBuilder::new().connect_provider(base_provider)`. This was implemented on the incorrect assumption that the fillers with which the `base_provider` was constructed would overwrite the "recommended fillers" introduced in `ProviderBuilder::new()`. 
This is not the case - the fillers in the `base_provider` are completely ignored, which I verified by trying to send transactions from a provider constructed with `ProviderBuilder::default().connect_provider(base_provider)` (no fillers are added in `default`). As such, this means the recommended fillers (notably, the `CachedNonceManager`) were used in the constructed provider.
We now construct an appropriately-configured provider in all cases where we need one, instead of trying to incorrectly extend some base provider.

Second, we employ the `send_tx_with_retry` helper when we `approve_erc20_allowance`. Even when fetching the nonce before sending a transaction, it's possible for it to be outdated by the time the transaction is sent.

### Testing
- [x] Run local funds manager, assert the following error reproduces before this fix and is resolved afterwards:
    1. Submit a swap through Lifi (sends a TX onchain)
    2. Deposit funds into Hyperliquid (uses the same private key, but different provider instance, incrementing the nonce in a way that a `CachedNonceManager` wouldn't detect)
    3. Submit another swap through Lifi (this would error before the fix, but now works smoothly)

### DIFF
--- a/funds-manager/funds-manager-server/src/cli.rs
+++ b/funds-manager/funds-manager-server/src/cli.rs
@@ -262,8 +262,7 @@ impl ChainConfig {
             &self.rpc_url,
             price_reporter.clone(),
             quoter_hot_wallet_private_key,
-        )
-        .map_err(FundsManagerError::custom)?;
+        );
 
         // Build a metrics recorder
         let metrics_recorder = MetricsRecorder::new(price_reporter.clone(), &self.rpc_url, chain);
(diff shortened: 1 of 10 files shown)


joeykraut (approved):
nice catch

PR #348, opened 2025-08-07
funds-manager: execution_client, cli: decay swap due to price deviation

akirillo:
This PR tweaks the price deviation check for execution quotes to downsize the quotes if it fails, rather than erroring early. Additionally, in the `ChainConfig` object, we now expect a map from token ticker to maximum price deviation allowed in quotes for that token.

### Testing
- [x] Ran local funds manager, set low deviation maximums, asserted that quotes are downsized when the maximum is exceeded


funds-manager/funds-manager-server/src/cli.rs
```diff
@@ -200,9 +204,12 @@ pub struct ChainConfig {
     /// is omitted
     pub protocol_decryption_key: Option<String>,
 
-    // --- Execution Venue Params --- //
+    // --- Execution Params --- //
     /// The Lifi API key
     pub lifi_api_key: Option<String>,
+    /// A map from token ticker to the maximum price deviation allowed in a
+    /// quote for that token
+    pub max_price_deviations: MaxPriceDeviations,
```
joeykraut, line 212: maybe default this with `#[serde(default)]` so that it is only optionally specified

PR #365, opened 2025-08-14
funds-manager: custody_client, handlers: swap for gas sponsor refill

akirillo:
This PR implements swapping into the tokens needed to refill the gas sponsor. We do this by first swapping into USDC, excluding all tokens which need to be refilled from the swap, and then swapping into the target tokens, so that we can maintain the same semantics around "buys," "sells," and "base" tokens in our post-swap telemetry.

### Testing
Deferred until we replace gas sponsor allocation, which greatly simplifies testing

### TODO
- Remove gas sponsor allocation in favor of fixed dollar value of sponsorship funds per token


funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -330,6 +385,64 @@ impl ExecutionClient {
         Ok(swap_outcome)
     }
 
+    /// Get the USDC value of the amount to purchase to cover the target balance
+    /// for each input token
```
joeykraut, line 389: This comment is a bit hard to read, do you mind rewording it to be clearer?

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -1,14 +1,14 @@
 //! Handlers for executing swaps
 
-use std::cmp::Ordering;
+use std::{cmp::Ordering, iter};
```
joeykraut, line 3: this file is large at this point, can you break it apart where you think is natural?

funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -429,7 +429,41 @@ pub(crate) async fn refill_gas_sponsor_handler(
...
+    let quoter_wallet = custody_client.get_quoter_hot_wallet().await?;
+    let signer = custody_client.get_hot_wallet_private_key(&quoter_wallet.address).await?;
+
+    let allocation = custody_client.fetch_gas_sponsor_allocation().await?;
+
+    // Refill the gas sponsor with native ETH
+    custody_client.refill_gas_sponsor_eth(&allocation).await?;
+
+    // Refill the gas sponsor with ERC20s
+
```
joeykraut, line 445: nit: too much whitespace

funds-manager/funds-manager-server/src/handlers.rs
```diff
@@ -429,7 +429,41 @@ pub(crate) async fn refill_gas_sponsor_handler(
     server: Arc<Server>,
```
joeykraut, line 429: this file is also very large, can you divide it up as well?

funds-manager/funds-manager-server/src/execution_client/swap.rs
```diff
@@ -330,6 +385,64 @@ impl ExecutionClient {
...
+        purchase_values: &[(Token, f64)],
+    ) -> Result<Vec<DecayingSwapOutcome>, ExecutionClientError> {
+        let usdc_token = Token::from_ticker_on_chain(USDC_TICKER, self.chain);
+
+        let usdc_target_balance = purchase_values.iter().map(|(_, value)| value).sum();
+        let exclude_tokens = purchase_values.iter().map(|(token, _)| token.get_addr()).collect();
+
+        let swap_request = SwapIntoTargetTokenRequest {
+            target_amount: usdc_target_balance,
+            quote_params: QuoteParams { from_token: usdc_token.get_addr(), ..Default::default() },
```
joeykraut, line 419: perhaps i'm reading this wrong, but how is UDSC the `from_token` here? Isn't this method intended to buy USDC

akirillo: indeed, this was a bug, fixed in the next PR. good catch

PR #381, opened 2025-08-19
auth-server: gas_sponsorship: retrieve refund amount by nonce

akirillo:
This PR adds the gas sponsorship `nonce` to the `BundleContext` so that we can ensure that the refund amount we fetch from contract logs lines up with the nonce used in the bundle's sponsorship info.

This is necessary because it's possible for the Darkpool to be traded through multiple times in one transaction, so we need to be sure we're fetching the refund amount from the correct log (not just the first one).

### Testing
- [x] Test Arbitrum / Base
- [x] Test buy / sell
- [x] Test in-kind / native sponsorship

### DIFF
--- a/auth/auth-server/src/server/api_handlers/settlement.rs
+++ b/auth/auth-server/src/server/api_handlers/settlement.rs
@@ -10,7 +10,6 @@ use crate::bundle_store::{BundleContext, helpers::generate_bundle_id};
 use crate::error::AuthServerError;
 use crate::server::api_handlers::external_match::SponsoredAssembleMalleableQuoteResponseCtx;
 use crate::telemetry::abi_helpers::{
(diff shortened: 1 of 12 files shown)


auth/auth-server/src/server/api_handlers/settlement.rs
```diff
@@ -29,6 +30,7 @@ impl Server {
         let resp = ctx.response();
         // Extract the nullifier from the original match bundle
         let nullifier = extract_nullifier_from_match_bundle(&resp.match_bundle)?;
+        let nonce = extract_nonce_from_match_bundle(&resp.match_bundle)?;
```
joeykraut, line 33: why are we decoding a value we just encoded to make this work? that seems silly and like a waste of compute.

We could place the gas sponsorship info in the request context if we like

joeykraut (approved):

PR #382, opened 2025-08-19
solver: fill orders with multiple outputs

sehyunc:
### Purpose
This PR enables the solver to handle orders with multiple output entries, as long as they are the same token. This is fairly trivial as the `Executor` must simply maintain the invariant that it holds sufficient output tokens by the time `reactorCallback` is finished. The `Reactor` contract is responsible for actually transferring these output tokens from the `Executor` to recipient(s).

### DIFF
--- a/renegade-solver/Cargo.toml
+++ b/renegade-solver/Cargo.toml
@@ -29,7 +29,7 @@ renegade-common = { workspace = true }
 renegade-config = { workspace = true }
 renegade-constants = { workspace = true }
 renegade-solidity-abi = { workspace = true }
-renegade-sdk = "0.1.15"
+renegade-sdk = { git = "https://github.com/renegade-fi/rust-sdk" }
 renegade-util = { workspace = true, features = [
     "telemetry",
 ] }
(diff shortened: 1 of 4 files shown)


renegade-solver/src/uniswapx/solve.rs
```diff
@@ -107,15 +109,20 @@ impl UniswapXSolver {
     ///
     /// TODO: Loosen and remove this method's checks in follow-ups
     fn temporary_order_filter(&self, order: &PriorityOrder) -> SolverResult<bool> {
-        // For now we only support single-leg routes
-        if order.outputs.len() != 1 {
-            return Ok(false);
+        // For now, we only support orders with 1 output token
```
joeykraut, line 112: nit: can you re-word this to say "the same token for all outputs"? I mis-read this as meaning "a single output"

renegade-solver/Cargo.toml
```diff
@@ -29,7 +29,7 @@ renegade-common = { workspace = true }
 renegade-config = { workspace = true }
 renegade-constants = { workspace = true }
 renegade-solidity-abi = { workspace = true }
-renegade-sdk = "0.1.15"
+renegade-sdk = { git = "https://github.com/renegade-fi/rust-sdk" }
```
joeykraut, line 32: want to just publish a new version of the `renegade-sdk`?

joeykraut (approved):

PR #385, opened 2025-08-23
funds-manager: manual awaiting of required confirmations

akirillo:
This PR implements manual awaiting of the desired number of confirmations after a transaction is included in a block. We do this because empirically, `alloy` will always immediately return a tx receipt as soon as the RPC provides it, ignoring the `required_confirmations`.

I'm investigating why this is the case - perhaps the block time of the chains we're on is too fast and is causing some race condition in the alloy internals. But until there is a resolution there, this approach lets us actually reap the `required_confirmations` guarantees we need.

This requires that we use a websocket provider throughout the funds manager, so we add the websocket RPC url to the chain configs.

### Testing
- [x] Ran local testnet funds manager, invoked withdrawals on base & arbitrum, confirmed that required confirmations were met.
- [x] Tested multiple concurrent interactions across wallets to ensure that using the same websocket connection did not result in any issues

### DIFF
--- a/funds-manager/funds-manager-server/src/helpers.rs
+++ b/funds-manager/funds-manager-server/src/helpers.rs
@@ -6,7 +6,7 @@ use std::{str::FromStr, time::Duration};
 use alloy::{
     providers::{
         fillers::{BlobGasFiller, ChainIdFiller, GasFiller},
-        DynProvider, Provider, ProviderBuilder,
+        DynProvider, Provider, ProviderBuilder, WsConnect,
     },
     rpc::types::{TransactionReceipt, TransactionRequest},
     signers::local::PrivateKeySigner,
@@ -19,6 +19,7 @@ use aws_config::SdkConfig;
 use aws_sdk_s3::Client as S3Client;
 use aws_sdk_secretsmanager::client::Client as SecretsManagerClient;
 use bigdecimal::{BigDecimal, FromPrimitive, RoundingMode, ToPrimitive};
+use futures::StreamExt;
 use rand::Rng;
 use renegade_common::types::chain::Chain;
 use renegade_util::{err_str, telemetry::helpers::backfill_trace_field};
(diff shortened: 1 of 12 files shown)


funds-manager/funds-manager-server/src/helpers.rs
```diff
@@ -78,23 +76,31 @@ sol! {
...
         .filler(ChainIdFiller::default())
         .filler(GasFiller)
         .filler(BlobGasFiller);
 
-    let provider = if let Some(wallet) = wallet {
-        builder.wallet(wallet).connect_http(url).erased()
+    if let Some(wallet) = wallet {
+        builder
+            .wallet(wallet)
+            .connect_ws(url)
```
joeykraut, line 93: I don't think we should be connecting a new websocket each time we fetch a provider. That's going to make everything slower and is just not practically a good implementation choice.

funds-manager/funds-manager-server/src/helpers.rs
```diff
@@ -120,6 +126,25 @@ pub async fn send_tx_with_retry(
         let mut pending_tx = pending_tx_res.map_err(FundsManagerError::on_chain)?;
         pending_tx.set_required_confirmations(required_confirmations);
         let receipt = pending_tx.get_receipt().await.map_err(FundsManagerError::on_chain)?;
+
```
joeykraut, line 129: this method is quite long now, can you break it up

funds-manager/funds-manager-server/src/helpers.rs
```diff
@@ -117,16 +121,47 @@ pub async fn send_tx_with_retry(
...
+    // TEMP: Alloy `required_confirmations` is empirically ignored. So for now,
+    // we manually watch blocks until the correct # of confirmations is reached.
+
+    let receipt_block =
+        receipt.block_number.ok_or(FundsManagerError::on_chain("No block number in receipt"))?;
+
+    // This method requires us to use a websocket provider
+    let block_subscription =
+        client.subscribe_blocks().await.map_err(FundsManagerError::on_chain)?;
```
joeykraut, line 153: Couple questions:
1. Will this re-subscribe and un-subscribe each time? I want to make sure that there's no way we're double streaming (or `n`-streaming) blocks. We get charged by bandwidth from alchemy
2. Is there a cheaper form of this that doesn't send the whole block? I.e. a cheaper subscription that receives less data?

Mostly aimed at not ballooning our costs. If these aren't possible that's okay since we're going to remove this when alloy fixes the bug.

joeykraut (approved):
looks great

PR #390, opened 2025-08-25
solver: add base fee cache

sehyunc:
### Purpose
This PR adds a base fee cache that is updated on every L2 block. We do this to remove retrieval of the latest base fee from the hot path of building a tx.

### DIFF
--- a/renegade-solver/src/fee_cache/fees.rs
+++ b/renegade-solver/src/fee_cache/fees.rs
@@ -2,9 +2,6 @@
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 
-/// The default value for the base fee per gas.
-const NONE: u64 = 0;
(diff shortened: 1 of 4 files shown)


renegade-solver/src/fee_cache/fees.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// A value-type handle to the fee cache. Clones are cheap and share state.
+#[derive(Clone)]
+pub struct FeeCache(Arc<FeeCacheInner>);
+
+impl Default for FeeCache {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
joeykraut, line 22: Any reason for implementing `Default` here? There's not a strong rule but it tends to be odd for `Default` to heap allocate -- which `Arc::new` will do

PR #398, opened 2025-08-26
funds-manager: execution_client: swap_immediate: introduce `SwapControlFlow` type

akirillo:
This PR introduces the `SwapControlFlow` type in the `swap_immediate` method, making it easier to break the core swap loop apart into multiple helpers instead of having to rely on `continue` and `return` statements within the loop body.

This will help keep the code clean as we implement more control flow options in the swap loop in the coming PRs.

### DIFF
--- a/funds-manager/funds-manager-server/src/execution_client/swap/swap_immediate.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/swap/swap_immediate.rs
@@ -35,6 +35,31 @@ const PRICE_DEVIATION_INCREASE: f64 = 0.2; // 20%
 /// allowed
 const MAX_PRICE_DEVIATION_INCREASE: f64 = 4.0; // 4x
 
+// ---------
+// | Types |
+// ---------
+
+/// The unsuccessful control flow branches that can occur during the execution
+/// of a decaying swap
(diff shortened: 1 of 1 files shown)


funds-manager/funds-manager-server/src/execution_client/swap/swap_immediate.rs
```diff
@@ -35,6 +35,31 @@ const PRICE_DEVIATION_INCREASE: f64 = 0.2; // 20%
...
+/// of a decaying swap
+#[derive(Debug, Clone)]
+enum SwapControlFlow {
+    /// Break out of the loop with no swap outcome
+    NoSwap,
+    /// Continue the loop with a higher price deviation tolerance
+    IncreasePriceDeviation,
+    /// Continue the loop with a smaller swap size.
+    /// Includes the gas cost incurred in the swap attempt.
+    DecreaseSwapSize(U256),
```
joeykraut, line 52: nit: Up to but, but you could make a named parameter for this to annotate it as gas cost:
```rust
DecreaseSwapSize {
    gas_cost: U256,
}
```

joeykraut (approved):
clean

PR #400, opened 2025-08-26
funds-manager: execution_client: bebop, lifi: handle gas estimation errors gracefully

akirillo:
This PR tweaks the error handling of the `BebopClient` & `LifiClient` `send_tx` method to treat failures to send the TX in the same way that it treats reverted TXs. This is necessary because we will be catching reverts during gas estimation implicitly in `send_tx`, and we want to signal to the swap loop that this merits a retry, rather than propagating an error all the way to the endpoint handler.

### DIFF
--- a/funds-manager/funds-manager-server/src/execution_client/venues/bebop/mod.rs
+++ b/funds-manager/funds-manager-server/src/execution_client/venues/bebop/mod.rs
@@ -297,21 +297,31 @@ impl ExecutionVenue for BebopClient {
 
         info!("Executing Bebop {} quote", bebop_execution_data.route_source);
 
-        let receipt = self.send_tx(tx).await?;
-        let gas_cost = get_gas_cost(&receipt);
-        let tx_hash = receipt.transaction_hash;
+        match self.send_tx(tx).await {
+            Ok(receipt) => {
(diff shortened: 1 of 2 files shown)


funds-manager/funds-manager-server/src/execution_client/venues/bebop/mod.rs
```diff
@@ -297,21 +297,31 @@ impl ExecutionVenue for BebopClient {
...
+
+                if receipt.status() {
+                    let recipient = bebop_execution_data.from;
+                    let buy_token_address = quote.buy_token.get_alloy_address();
+                    let buy_amount_actual =
+                        get_received_amount(&receipt, buy_token_address, recipient);
+
+                    Ok(ExecutionResult { buy_amount_actual, gas_cost, tx_hash: Some(tx_hash) })
+                } else {
+                    warn!("tx ({:#x}) reverted", tx_hash);
```
joeykraut, line 313: nit: inline the `tx_hash` into the format, this is a clippy error in rust 2024. I.e.
```rust
warn!("tx ({tx_hash:#x}) reverted");
```

Same in the other venue

joeykraut (approved):
if we add more venues, we should probably make this behavior generic in some way.

PR #406, opened 2025-08-30
solver: build tx before awaiting send

sehyunc:
### Purpose
This PR modifies the solver to store a ready-to-send tx in the TxStore, such that when the time to submit a tx arrives there is no work to be done except submission. The benefits of this are two-fold:
- hot path of tx submission is optimized
- we know the tx hash before submission ensuring we record confirmation. Previously, there was a race between the sequencer and the flashblocks listener as sometimes the sequencer would be slow to respond with the tx hash, resulting in inclusion in a flashblock being missed.

### DIFF
--- a/renegade-solver/src/fee_cache/fees.rs
+++ b/renegade-solver/src/fee_cache/fees.rs
@@ -6,6 +6,8 @@ use std::sync::Arc;
 struct FeeCacheInner {
     /// The base fee per gas.
     base_fee_per_gas: AtomicU64,
+    /// The pending nonce for the signer address.
+    pending_nonce: AtomicU64,
 }
 
 #[derive(Clone)]
@@ -15,7 +17,10 @@ pub struct FeeCache(Arc<FeeCacheInner>);
 impl FeeCache {
     /// Create a new cache
     pub fn new() -> Self {
-        Self(Arc::new(FeeCacheInner { base_fee_per_gas: AtomicU64::default() }))
+        Self(Arc::new(FeeCacheInner {
+            base_fee_per_gas: AtomicU64::default(),
+            pending_nonce: AtomicU64::default(),
+        }))
     }
 
     /// Sets the base fee per gas.
@@ -26,6 +31,19 @@ impl FeeCache {
     /// Gets the base fee per gas.
     pub fn base_fee_per_gas(&self) -> Option<u64> {
         match self.0.base_fee_per_gas.load(Ordering::Relaxed) {
+            0 => None,
+            v => Some(v),
+        }
+    }
+
+    /// Sets the pending nonce for the signer address.
+    pub fn set_pending_nonce(&self, nonce: u64) {
+        self.0.pending_nonce.store(nonce, Ordering::Relaxed);
+    }
+
+    /// Gets the pending nonce for the signer address.
+    pub fn pending_nonce(&self) -> Option<u64> {
+        match self.0.pending_nonce.load(Ordering::Relaxed) {
             0 => None,
             v => Some(v),
         }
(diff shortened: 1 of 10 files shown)


renegade-solver/src/fee_cache/fees.rs
```diff
@@ -6,6 +6,8 @@ use std::sync::Arc;
 struct FeeCacheInner {
     /// The base fee per gas.
     base_fee_per_gas: AtomicU64,
+    /// The pending nonce for the signer address.
```
joeykraut, line 9: nit: is the fee cache the best place for this? perhaps if so we should generalize the name to `ChainStateCache` or similar

renegade-solver/src/fee_cache/worker.rs
```diff
@@ -1,34 +1,45 @@
...
+use crate::{cli::Cli, fee_cache::fees::FeeCache};
 
 /// The worker that listens for blocks and updates the fee cache.
 pub struct FeeCacheWorker {
     /// The provider to use to subscribe to blocks.
     provider: DynProvider,
     /// The fee cache to update.
     fee_cache: FeeCache,
+    /// Signer address whose pending nonce is tracked
+    signer_address: Address,
```
joeykraut, line 21: Probably should just colocate this with the pending nonce for the address. Doesn't really make sense to have them in different structs

renegade-solver/src/fee_cache/worker.rs
```diff
@@ -37,6 +48,11 @@ impl FeeCacheWorker {
                     if let Some(base) = header.base_fee_per_gas {
                         fee_cache.set_base_fee_per_gas(base);
                     }
+                    // Update pending nonce on each block tick
+                    match provider.get_transaction_count(signer).await {
```
joeykraut, line 52: do we absolutely need to be fetching the txn count each block? 

With 200ms flashblocks, this will be `1 fetch / 200 ms * 1000 ms / 1 s * 60 s / 1 m * 60 m / 1 hour * 24 hr / 1d = 432,000 req/day`. At 20 CUs per call to `eth_getTransactionCount`; this line of code alone will account for > 10% of our alchemy budget.

renegade-solver/src/tx_store/store.rs
```diff
@@ -28,14 +25,17 @@ pub struct L2Position {
...
-    /// subtracting the given number of flashblocks, respecting the block
-    /// boundaries.
-    pub fn sub_flashblocks(&self, flashblocks: u64, flashblocks_per_block: u64) -> Self {
-        let idx = self.l2_block * flashblocks_per_block + self.flashblock;
-        let idx = idx.saturating_sub(flashblocks);
-        let l2_block = idx / flashblocks_per_block;
-        let flashblock = idx % flashblocks_per_block;
+    /// Returns the linearized flashblock index for this position given the
+    /// flashblocks-per-block value.
+    pub fn linear_index(&self, flashblocks_per_block: u64) -> u128 {
```
joeykraut, line 30: In what sense is this linear? can you either document or choose a better name?

renegade-solver/src/uniswapx/solve.rs
```diff
@@ -70,20 +62,69 @@ impl UniswapXSolver {
...
-        let plan = compute_send_plan(target, &Measurements::default());
-        let timing: TxTiming = plan.into();
-        self.tx_store.enqueue_with_timing(&order_hash, tx, timing.clone())?;
+        Ok(())
+    }
 
-        self.log_writing_tx_record(&order_hash, priority_fee_wei, &timing);
+    /// Hydrate a transaction request with the latest base fee and nonce from
+    /// the cache
+    fn hydrate_tx_request(
```
joeykraut, line 104: "Hydrate" is empty, maybe `set_tx_fee`?

joeykraut (approved):

renegade-solver/src/fee_cache/worker.rs
sehyunc: This worker subscribes to L2 blocks (which should be made clearer) meaning it would be `43,200 req/day`, which is still a lot. I will change this to either update cache less frequently or fetch in the hot path.

PR #410, opened 2025-09-02
auth-server: bundle-store: Use nonce as bundle ID

joeykraut:
### Purpose
This PR uses the gas sponsorship nonce as the bundle ID and listens for `NonceUsed` events on-chain as opposed to nullifier spent events.

### Testing
- [x] Tested locally
- [ ] Tested in testnet


auth/auth-server/src/bundle_store/mod.rs
```diff
@@ -1,23 +1,23 @@
...
-use std::{
-    collections::{HashMap, HashSet},
-    sync::Arc,
-};
 
 use alloy_primitives::U256;
 use auth_server_api::GasSponsorshipInfo;
-use renegade_circuit_types::wallet::Nullifier;
-use tokio::sync::RwLock;
+use dashmap::DashMap;
```
akirillo, line 5: TIL about dashmap, neat

akirillo (approved):
very clean

PR #411, opened 2025-09-03
solver: add flashblock clock

sehyunc:
### Purpose
This PR adds the FlashblockClock which is reponsible for converting a (flashblock index, L2 block index) into a timestamp (ms since epoch). This timestamp is used as the target timestamp used by the arrival controller. It estimates flashblock and L2 block durations using EMAs and uses them to estimate when a future flashblock will be seen over the WebSocket.

### DIFF
--- a/renegade-solver/src/arrival_control/controller.rs
+++ b/renegade-solver/src/arrival_control/controller.rs
@@ -26,13 +26,13 @@ pub const INITIAL_DELAY_SEED_MS: f64 = 200.0;
 #[derive(Clone)]
 pub struct ArrivalController {
     /// The EMA instance for tracking delay estimates.
-    delay_ema: Arc<Mutex<Ema>>,
+    delay_ema: Arc<Ema>,
 }
 
 impl Default for ArrivalController {
     fn default() -> Self {
         let delay_ema = Ema::from_window_length(DELAY_WINDOW, INITIAL_DELAY_SEED_MS);
-        Self { delay_ema: Arc::new(Mutex::new(delay_ema)) }
+        Self { delay_ema: Arc::new(delay_ema) }
     }
 }
 
@@ -40,7 +40,7 @@ impl ArrivalController {
     /// Compute local timestamp at which to send to target an arrival at
     /// `target_ms`.
     pub fn compute_send_ms(&self, target_ms: u64) -> u64 {
-        let delay_estimate = self.delay_ema.lock().expect("EMA lock poisoned").last();
+        let delay_estimate = self.delay_ema.last();
         target_ms.saturating_sub(delay_estimate.round() as u64)
     }
 
@@ -54,7 +54,7 @@ impl ArrivalController {
     /// We approximate the one-way delay as the time between sending and
     /// observing the packet arrival.
     fn update_delay_estimate(&self, send_ms: u64, actual_ms: u64) {
-        let ema = self.delay_ema.lock().expect("EMA lock poisoned");
+        let ema = self.delay_ema;
         info!("old delay estimate: {}ms", ema.last());
 
         let observed_one_way_delay_ms = actual_ms.saturating_sub(send_ms) as f64;
--- a/renegade-solver/src/flashblocks/clock/ema.rs
+++ b/renegade-solver/src/flashblocks/clock/ema.rs
@@ -25,55 +25,29 @@ impl EmaManager {
     }
 
     /// Returns the current EMA estimate of the flashblock duration.
+    ///
+    /// Note: this rounds to the nearest millisecond.
     pub fn flashblock_duration_ms(&self) -> u64 {
         let est = self.fb.last();
         est.round() as u64
     }
 
     /// Returns the current EMA estimate of the L2 block duration.
+    ///
+    /// Note: this rounds to the nearest millisecond.
     pub fn l2_block_duration_ms(&self) -> u64 {
         let est = self.l2.last();
         est.round() as u64
     }
 
-    #[allow(clippy::too_many_arguments)]
-    /// Updates the EMA estimates of the flashblock and L2 block durations.
-    pub fn update_estimates(
-        &self,
-        last_flashblock_idx: u64,
-        last_flashblock_ts: u64,
-        last_l2_idx: u64,
-        last_l2_ts: u64,
-        current_flashblock_idx: u64,
-        current_l2_idx: u64,
-        now_ms: u64,
-    ) {
-        // Only update if we have valid samples
-        if let Some(fb_sample) = self.try_sample_flashblock(
-            last_flashblock_idx,
-            last_flashblock_ts,
-            current_flashblock_idx,
-            now_ms,
-        ) {
-            self.update_fb_estimate(fb_sample);
-        }
-
-        if let Some(l2_sample) = self.try_sample_l2(last_l2_idx, last_l2_ts, current_l2_idx, now_ms)
-        {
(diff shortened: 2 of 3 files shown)


renegade-solver/src/arrival_control/controller.rs
```diff
@@ -17,6 +19,9 @@ use crate::arrival_control::ema::Ema;
 /// noise.
 pub const DELAY_WINDOW: u32 = 12;
 
+/// Initial seed value (ms) for the delay EMA.
+pub const INITIAL_DELAY_SEED_MS: f64 = 200.0;
+
 /// A thread-safe arrival controller.
 #[derive(Clone)]
 pub struct ArrivalController {
```
joeykraut, line 27: You can remove the mutex here, `Ema` is already thread safe once you wrap it in an `Arc` because it's using atomics

renegade-solver/src/arrival_control/ema.rs
```diff
@@ -1,57 +1,52 @@
...
 /// The EMA
-#[derive(Clone, Copy, Debug)]
+#[derive(Debug)]
 pub struct Ema {
     /// The smoothing factor for the EMA.
     alpha: f64,
-    /// The current EMA estimate.
-    last: Option<f64>,
+    /// The current EMA estimate (seeded upfront; non-optional).
+    last: AtomicF64,
```
joeykraut, line 13: good call, probably not that important for performance sake, but much more ergonomic

renegade-solver/src/flashblocks/clock/ema.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn update_estimates(
+        &self,
+        last_flashblock_idx: u64,
+        last_flashblock_ts: u64,
+        last_l2_idx: u64,
+        last_l2_ts: u64,
+        current_flashblock_idx: u64,
+        current_l2_idx: u64,
+        now_ms: u64,
+    ) {
```
joeykraut, line 50: This is a lot of args of the same type in a row. That's generally a bad pattern because it's easy to introduce bugs by switching up args order. In some languages, linters will actually disallow this. Probably best to either:
1. Break up the method into multiple methods, one for each of the estimates we're updating and have the caller call out to both.
2. Define a structured type representing the args to this method so that they're named. Something like `UpdatedClockEstimates`.

Also I don't think we need `now_ms` as an arg, I'd just compute it in this method.

renegade-solver/src/flashblocks/clock/ema.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn update_fb_estimate(&self, sample_ms: u64) -> u64 {
+        let new = self.fb.update(sample_ms as f64);
+        new.max(1.0).round() as u64
+    }
+
+    /// Updates the EMA estimate of the L2 block duration.
+    fn update_l2_estimate(&self, sample_ms: u64) -> u64 {
+        let new = self.l2.update(sample_ms as f64);
+        new.max(1.0).round() as u64
+    }
```
joeykraut, line 77: Might want to note in a comment that we discretize into a ms granularity

renegade-solver/src/flashblocks/clock/ema.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// The flashblock is valid if:
+    /// - The previous flashblock index and timestamp are valid.
+    /// - The current flashblock index is greater than the previous flashblock
+    ///   index.
+    /// - The current flashblock index is not zero. This is because the first
+    ///   flashblock is special in that it does not contain user txns, so we
+    ///   ignore it.
+    ///
+    /// Returns the sample if the flashblock is valid.
+    fn try_sample_flashblock(
```
joeykraut, line 90: It's not clear from the name that this is returning a time / block value. It's _somewhat_ clear from context, but maybe you could name it `maybe_sample_flashblock_duration` or something similar?

Also I'd internalize `now_ms` or otherwise rename it to `current_flashblock_ts` to match the naming.

renegade-solver/src/flashblocks/clock/ema.rs
```diff
@@ -0,0 +1,136 @@
...
+        last_flashblock_idx: u64,
+        last_flashblock_ts: u64,
+        current_flashblock_idx: u64,
+        now_ms: u64,
+    ) -> Option<u64> {
+        if last_flashblock_idx == 0 || last_flashblock_ts == 0 || current_flashblock_idx == 0 {
+            return None;
+        }
+
+        let delta = current_flashblock_idx.saturating_sub(last_flashblock_idx);
```
joeykraut, line 101: nit: `idx_delta` is a slightly better name for an unfamiliar reader, and `ts_delta` below might help

renegade-solver/src/flashblocks/clock/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+        let last_l2_idx = self.0.last_l2_idx.load(Ordering::Relaxed);
+        if current_l2_idx != last_l2_idx {
+            self.0.last_l2_idx.store(current_l2_idx, Ordering::Relaxed);
+            self.0.last_l2_ts.store(now, Ordering::Relaxed);
+        }
+    }
+
+    /// Snapshot the current state of the clock.
+    pub fn snapshot(&self) -> FlashblockClockSnapshot {
+        FlashblockClockSnapshot {
```
joeykraut, line 116: style nit: i'd pull a reference to `inner` and `ema` up front to avoid all this `self.0.ema` business:
```rust
let inner = &self.0;
let ema = &inner.ema;
FlashblockClockSnapshot {
    flashblock_duration_ms: ema.flashblock_duration_ms(),
    ...
}
...
```

renegade-solver/src/flashblocks/clock/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+        let current_fb_idx = fb.index;
+        let current_l2_idx = fb.metadata.block_number;
+
+        let FlashblockClockSnapshot {
+            last_l2_idx,
+            last_l2_ts,
+            last_flashblock_idx,
+            last_flashblock_ts,
+            flashblock_duration_ms: _,
+            l2_block_duration_ms: _,
```
joeykraut, line 144: you can do `..` here for unbound fields:
```rust
let FlashblockClockSnapshot {
            last_l2_idx,
            last_l2_ts,
            last_flashblock_idx,
            last_flashblock_ts,
            ..
        } = self.snapshot();
```

renegade-solver/src/flashblocks/clock/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+            last_flashblock_idx: self.0.last_flashblock_idx.load(Ordering::Relaxed),
+            last_flashblock_ts: self.0.last_flashblock_ts.load(Ordering::Relaxed),
+        }
+    }
+}
+
+/// Returns the current unix timestamp in milliseconds, represented as u64.
+pub fn get_current_time_millis() -> u64 {
+    SystemTime::now().duration_since(UNIX_EPOCH).expect("negative timestamp").as_millis() as u64
+}
```
joeykraut, line 130: you have `renegade-util` already in scope; I'd use the method already defined [there](https://github.com/renegade-fi/renegade/blob/11c541e88a5fdc3938a141ebc959dbc5e97630d0/util/src/lib.rs#L34-L37)

joeykraut (approved):

PR #412, opened 2025-09-03
solver: use flashclock and controller to enqueue

sehyunc:
### Purpose
This PR hooks up the new `FlashblockClock` and `ArrivalController` in the solver to complete the new flow: 
1. Find solution
2. Sign txn
3. Estimate timestamp of target (flashblock, L2 block) using `FlashblockClock`
4. Enqueue directly onto the `TxDriver`
6. `TxDriver` `sleep`s until specified send timestamp then submits
7. Inclusion in a flashblock is observed in `chain-events-listener` and the controller updates on the observed delay.


joeykraut (commented):
🚀

joeykraut (approved):

PR #416, opened 2025-09-06
solver: update on actual submit latency

sehyunc:
### Purpose
This PR changes how we update our estimate the RTT between the solver and the block builder. Previously, we were using the timestamp at which we observe inclusion in a block as `actual_ts`; this captures more than just the latency regarding submitting a transaction since it includes the block building process and any lag when streaming flashblock data over the Websocket. Now, we exactly measure the timestamps before and after sending the `eth_sendRawTransaction` RPC as the submission latency and update on that.

### DIFF
--- a/renegade-solver/src/tx_driver/driver.rs
+++ b/renegade-solver/src/tx_driver/driver.rs
@@ -1,52 +1,91 @@
 //! Defines the transaction driver which schedules submissions by timestamp
 //! millis.
 
-use alloy_primitives::{Bytes, TxHash};
+use alloy_primitives::Bytes;
 use eyre::Result;
 use renegade_util::get_current_time_millis;
 use tokio::time::{sleep_until, Duration, Instant};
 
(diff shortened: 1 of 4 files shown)


renegade-solver/src/tx_driver/driver.rs
```diff
@@ -1,52 +1,91 @@
...
+        arrival_controller: ArrivalController,
+        tx_store: TxStore,
     ) -> Result<()> {
-        sleep_until(send_at).await;
-        executor_client.send_raw(raw_tx_bytes).await?;
+        // Sleep until the send time
+        let now = get_current_time_millis();
+        let delay_ms = send_ts.saturating_sub(now);
+        let deadline = Instant::now() + Duration::from_millis(delay_ms);
+        sleep_until(deadline).await;
```
joeykraut, line 54: maybe just add a `sleep_until_timestamp` helper?

joeykraut (approved):

PR #419, opened 2025-09-09
auth-server: scaffold `/rfqt/v3/levels`

sehyunc:
### Purpose
This PR scaffolds the `/rfqt/v3/levels` endpoint with a response with the correct shape but dummy values.

### DIFF
--- a/auth/auth-server/src/main.rs
+++ b/auth/auth-server/src/main.rs
@@ -412,10 +412,7 @@ async fn main() {
             server.handle_order_book_request(path, headers).await
         });
 
-    let rfqt_levels_path = warp::path("rfqt")
-        .and(warp::path("v3"))
-        .and(warp::path("levels"))
-        .and(warp::path::end())
+    let rfqt_levels_path = warp::path!("rfqt" / "v3" / "levels")
         .and(warp::path::full())
         .and(warp::header::headers_cloned())
         .and(with_query_string())
--- a/auth/auth-server/src/server/api_handlers/rfqt/helpers.rs
+++ b/auth/auth-server/src/server/api_handlers/rfqt/helpers.rs
@@ -11,7 +11,6 @@ pub fn parse_levels_query_params(
     query_str: &str,
     server_chain: Chain,
 ) -> Result<RfqtLevelsQueryParams, AuthServerError> {
-    // If string is empty, return default (do nothing)
     if query_str.is_empty() {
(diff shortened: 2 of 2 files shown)


auth/auth-server/src/server/api_handlers/rfqt/types.rs
```diff
@@ -0,0 +1,78 @@
+//! RFQT endpoint types
```
joeykraut, line 1: these should be in the `auth-server-api` crate colocated with the `auth-server`. It's nice to have them in a separate crate which a client can import directly.

auth/auth-server/src/server/api_handlers/rfqt/types.rs
```diff
@@ -0,0 +1,78 @@
...
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+/// Query params for GET /rfqt/v3/levels
+#[derive(Debug, Deserialize, Default)]
+pub struct RfqtLevelsQueryParams {
+    /// Chain identifier (1=Ethereum, 137=Polygon, 42161=Arbitrum)
+    #[serde(rename = "chainId")]
+    pub chain_id: Option<String>,
```
joeykraut, line 11: I'd make this an enum and only accept supported chains

auth/auth-server/src/main.rs
```diff
@@ -412,6 +412,19 @@ async fn main() {
             server.handle_order_book_request(path, headers).await
         });
 
+    let rfqt_levels_path = warp::path("rfqt")
+        .and(warp::path("v3"))
+        .and(warp::path("levels"))
```
joeykraut, line 417: use [`warp::path`](https://docs.rs/warp/latest/warp/macro.path.html) here

joeykraut (approved):

auth/auth-server/src/server/api_handlers/rfqt/types.rs
sehyunc: Not sure this is necessary since we have a different subdomain for each chain the query parameter is kind of useless. I'm assuming 0x can handle this, if not we may need another server.

sehyunc: Oh yes, I mean not sure and enum is necessary. I've added validation, that was a good call out.

PR #441, opened 2025-10-07
coinbase websocket tweaks

akirillo:


price-reporter/src/price_stream_manager.rs
```diff
@@ -88,7 +83,22 @@ impl GlobalPriceStreams {
...
+            // tasks being spawned.
+
+            let mut price_streams = self.price_streams.write().await;
+
+            if let Some(stream_rx) = price_streams.get(&pair_info).cloned() {
+                return Ok(stream_rx);
+            }
+
+            price_streams.insert(pair_info.clone(), price_rx.clone());
+        }
```
joeykraut, line 101: cleaner to encapsulate this into a function which will naturally drop the RAII guard; rather than a large scoped block in the middle of your method here. maybe something like `self.maybe_add_price_stream`?

joeykraut (approved):

